PR #5, opened 2021-04-20
Solve missing quantum.h on WS2812 driver for AVR

jkutianski:
Solve missing quantum.h on WS2812 driver for AVR mC


jkutianski:
To PR on QMK

PR #13, opened 2021-05-18
raw_hid_receive_kb for unhandled command_ids

JanLunge:
this adds the layer switching functions like in qmk_rc to a vial supported keyboard
I added this so that a host application can switch between layers on the keyboard for specific programs like custom hotkeys while video editing.

the keyboard config for my 3by2 macropad is added as well


JanLunge:
to be more precise the raw_hid_recieve function did not pass unhandled codes to the raw_hid_recieve_kb so raw_hid_recieve was not possible to be implemented on a keyboard level this is what this pr tries to resolve

xyzz:
merged in afb490c2977d68fbf826db7f2b153bba37a6c673 and edd71157666ece4af764bc913b170cc9b787f619

PR #20, opened 2021-06-18
Update rules.mk

cubeydice:
This PR enables on the fly rotary encoder support in Vial. Allows users to add Vial encoders as part of their KLE keymap
[Vial Encoder Support](https://get.vial.today/gettingStarted/encoders.html)


xyzz:
I don't see the actual vial keymap here so how's that supposed to be used?

PR #34, opened 2021-07-18
Vial support for MechWild Mercutio

kylemccreery:
Added Vial keymap for MechWild Mercutio.


keyboards/mechwild/mercutio/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,5 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+VIAL_ENCODERS_ENABLE = yes
+
+EXTRAFLAGS += -flto
```
xyzz: any reason to use this instead of `LTO_ENABLE = yes`?

kylemccreery: Changed it. It was just the legacy LTO blurb that was included when I first wrote this firmware and I didn't change it for this update.

PR #36, opened 2021-07-20
Pteron36 Oled, rgb support, enable ee_hand

harshitgoel96:
Enable RGB, Oled, and EE_HAND


harshitgoel96:
my build command works fine.. no issues in building via keymap. the keymap is at 97% usage

PR #37, opened 2021-07-23
Vial support for MechWild OBE

kylemccreery:
Added Vial keymap for MechWild OrangeBoyErgo (OBE).


xyzz:
fyi - it's recommended to use Vial's built-in [stm32f4 eeprom emulation driver](https://get.vial.today/changelog/release-0.4.html#stm32f4-eeprom-emulation-support), as opposed to i2c, due to better performance (much lower latency since it doesn't need an i2c transaction for reading every byte - and existing VIA code is quite inefficient with how it's reading eeprom data), balanced wear leveling, and most importantly not requiring any additional hardware

You're welcome to try it, and feel free to update the keyboard if you find the implementation works well for your use case.

kylemccreery:
> fyi - it's recommended to use Vial's built-in [stm32f4 eeprom emulation driver](https://get.vial.today/changelog/release-0.4.html#stm32f4-eeprom-emulation-support), as opposed to i2c, due to better performance (much lower latency since it doesn't need an i2c transaction for reading every byte - and existing VIA code is quite inefficient with how it's reading eeprom data), balanced wear leveling, and most importantly not requiring any additional hardware
> 
> You're welcome to try it, and feel free to update the keyboard if you find the implementation works well for your use case.

@xyzz Trying it out now. Looks like it works for everything except saving the LED state. Not sure the specifics on that, but it reverts on reset.

xyzz:
Do you mean rgblight configuration? are you actually pressing the "save" button in Vial GUI? I'm not sure why only that one wouldn't work.

kylemccreery:
Yes. I used the save button. It doesn't affect the issue. I tried changing
the LED through VIA and using keycodes as well. The LED status is not saved
under any of those methods.

xyzz:
I think there's some race condition in rgblight implementation, when disabling RGBLIGHT_SLEEP it actually restores the lighting for me. In fact, it doesn't seem `rgblight_wakeup` gets called... and also adding a sleep into rgblight_init fixes the issue - leading me to suspect a race condition

xyzz:
Another way to see this issue would be to set `EEPROM_DRIVER = transient` (meaning it's just kept in memory and reset every boot), ensure you also have `#define TRANSIENT_EEPROM_SIZE 4096`. Then, you would expect a solid red color on every boot as eeprom contents aren't preserved and rgblight is reset, however it is instead turned off.

kylemccreery:
Interesting. This only happens with me currently from the emulation
implemented in Vial so I am unable to dig into it with other settings too
much.

PR #43, opened 2021-08-18
My new macropad

Ll3ma:
its my new macropad called The PASSWORD my keymap is called vial and i will be releasing most of the details soon


xyzz:
please add a "default" keymap so that the build can pass

Ll3ma:
> please add a "default" keymap so that the build can pass

fixed

xyzz:
there should be two keymaps, the "default" keymaps shouldn't have via or vial enabled, and the "vial" keymaps should have it enabled

Ll3ma:
fixed

PR #50, opened 2021-08-29
add vial support for Sofle and its encoders

kirbysayshi:
- Took Via JSON and used it to create vial.json
- Created KLE mapping for default sofle rotary encoders ( <-- this is why I did this!)
- Added unlock combo of four corners of the board
- enabled mousekeys, disabled QMK_SETTINGS and rgb support to ensure small enough firmware

This is the first time I've ever done any work with QMK+Vial, so any feedback is encouraged!


kirbysayshi:
Initial failure was due to clang-format formatting the UID on a separate line

PR #56, opened 2021-09-12
Add vial support for Keycapsss' Kimiko

ykz89:
Add support for Keycapsss' Kimiko.
To be able to compile the firmware rgb and encoder support are disabled, currently at  (`28626/28672 (99%, 46 bytes free)`).
Unlock combo is set to Escape+Enter as per the documentation.


xyzz:
the firmware size is too large so the build is failing, could you reduce the size:

```
2021-09-12T16:52:55.4288141Z Checking file size of keycapsss_kimiko_rev1_vial.hex                                               
2021-09-12T16:52:55.4289379Z  * [31;01mThe firmware is too large![0m 29060/28672 (388 bytes over)
2021-09-12T16:52:55.4295985Z  [31;01m[ERRORS][0m
```

PR #63, opened 2021-10-24
Initial commit, adding full support for GMMK Pro ANSI

usrfriendly:
Adds full support for GMMK Pro ANSI layout, including RGB backlight configuration via Vial.  Backlight changing has been tested to work.


xyzz:
Looks like the build is failing with `UID 81d28a891736534b duplicated: {'gmmk/pro', 'gmmk/pro/ansi'}`, I think the way it should work is that the current vial keymap in `gmmk/pro` should be removed, then the new config put in `gmmk/pro/ansi` to specify unique keyboard ID. So this way `gmmk/pro/ansi` and `gmmk/pro/iso` could have different Vial ID (though no need to touch the iso version in this pr if you don't have it)

usrfriendly:
Build failed again because of redundant `#VIAL_KEYBOARD_UID` entries.  It even failed on my WSL/Debian setup.  Removing the entry from either pro/config.h or pro/ansi/config.h allows it to build on my setup, but it won't pass the UID check when building if I push it that way.

xyzz:
Yeah, actually my suggestion would be to remove the `gmmk/pro/keymaps` altogether, since it sees `gmmk/pro/keymaps/via` and tries to compile it, and that keymap isn't even in qmk anymore after they split it into ansi/iso subfolders.

Additionally, you shouldn't touch ` keyboards/gmmk/pro/ansi/config.h` but instead put it into keymap `config.h`, i.e. ` keyboards/gmmk/pro/ansi/keymaps/vial/config.h`, this way any change within qmk to the root config.h would propagate cleanly (I assume that's also the reason why it kept the old `via` keymap under `keyboards/gmmk/pro`, while main qmk doesn't have it anymore after the keyboard's been split into ansi/iso)

usrfriendly:
Thanks for the help!  I just commented out the UID in pro/config.h if that's alright, just in case we wound up needing it, but it passed with the UID in ansi/keymaps/vial/config.h

PR #65, opened 2021-10-29
Added gmmk/pro/iso files for full rotary knob support

nono-was-taken:
* I took the ../ansi/keymaps/vial folder and edited all files for (hopefully) full ISO support.


xyzz:
since it's a different PCB, it should have a different VIAL_KEYBOARD_UID

nono-was-taken:
you're right - totally forgot that. should be fixed by now.

PR #72, opened 2021-11-08
VS Code settings update

310weber:
Updated settings for VS Code to enable MSYS terminal for easy compiling.


310weber:
Sorry, created in error.

310weber:
Updated settings for VS Code to enable MSYS terminal for easy compiling.

PR #80, opened 2021-11-15
Added support for Keebio BDN9 and Cannon Keys db60

ambustion:


keyboards/cannonkeys/db60/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,4 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+TAP_DANCE_ENABLE = yes
+VIAL_INSECURE = yes
```
xyzz: should be removed

keyboards/keebio/bdn9/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,9 @@
+VIA_ENABLE = yes
+LTO_ENABLE = yes
+MOUSEKEY_ENABLE = yes
+CONSOLE_ENABLE = yes
+COMMAND_ENABLE = no
+VIAL_ENABLE = yes
+VIAL_ENCODERS_ENABLE = yes
+ENCODER_ENABLE = yes
+VIAL_INSECURE = yes
```
xyzz: same as above, should be removed

xyzz (changes requested):

xyzz:
- db60/keymaps/vial/rules.mk  is empty, vial isn't actually getting enabled
- db60/keymaps/vial/db60.jsonZone.Identifier shouldn't be in the pr
- db60/keymaps/vial/db60.json also shouldn't be in the pr

keyboards/keebio/bdn9/keymaps/vial/rules.mk
ambustion: Is it the insecure part or the lines for encoders?

xyzz:
the build is still failing

keyboards/keebio/bdn9/keymaps/vial/keymap.c
```diff
@@ -0,0 +1,67 @@
...
+
+[_THREE] = LAYOUT(
+    _______, _______, _______,
+    _______, _______, _______,
+    _______, _______, _______
+)
+
+};
+
+bool encoder_update_user(uint8_t index, bool clockwise) {
```
xyzz: any reason to have this here when you're using vial encoders?

keyboards/keebio/bdn9/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,11 @@
+LTO_ENABLE = yes
+COMMAND_ENABLE = no
+ifdef RGB_MATRIX_ENABLE
```
xyzz: While this is a valid syntax i really don't understand what you're trying to do here; e.g. why are you only enabling Vial when RGB matrix is enabled?

xyzz (changes requested):

keyboards/keebio/bdn9/keymaps/vial/rules.mk
ambustion: The build fails on the BDN9 rev1 due to firmware size so I used that as a check to enable only for the rev2. If there's a way to check for which rev is being built that would be more correct but I'm not sure how to do that.

keyboards/keebio/bdn9/keymaps/vial/keymap.c
ambustion: No other than it was in the default keymap I was using.

keyboards/keebio/bdn9/keymaps/vial/rules.mk
xyzz: If that's the case, can you put the keymap under `keebio/bdn9/rev2/keymaps/vial`? Otherwise, the build bot isn't even currently trying to build rev2 so it's kinda pointless to have vial support for it in vial-qmk.

ambustion:
Ok alright, I didn't realize that path structure was ok. Thank you for your
guidance.

Brendon Rathbone
C: 403.305.4443

PR #81, opened 2021-11-16
Reviung34 multiple layouts and rgb support

kbrock:
The reviung 34 has options for 2x1u or 1x2u in the space bar position.
This gives users the option to view as a single or multiple.

Some models of the reviung 34 have rgb support, specifically the unibody version.
I added support but the memory was pushing it so I only enabled one animation.

If you would prefer I split this up into 2 PRs, let me know

/cc @gtips thanks for the great boards and give a shout if you would prefer I implement this a different way.

My new kle is http://www.keyboard-layout-editor.com/#/gists/ce988fc2da8ab7439df374cecdfb5c6e if you want to merge those changes into your version


gtips:
Thanks!

PR #83, opened 2021-11-19
bongopad macropad

Ll3ma:
this is one of my new macropads


Ll3ma:
les go

xyzz:
- VIAL_INSECURE shouldn't be set
- the vial keymap should be named "vial"
- there are unrelated submodule changes under lib/lufa

PR #85, opened 2021-11-23
adding bongopad

Ll3ma:
updated


xyzz:
Looks like it's failing to build, can you reduce firmware size

ll3macorn:
> Looks like it's failing to build, can you reduce firmware size

its not the bongopad failing to built thats for sure

PR #87, opened 2021-12-03
Add vial support the murcielago

tomsquest:
The Murcielago  is already supported in Qmk and Via.

- `vial.json` taken from https://github.com/the-via/keyboards/blob/master/src/elagil/murcielago/murcielago.json
- `LTO_ENABLE = yes` added to reduce the size (~97% with the option, so required).

Pinging @elagil, the creator, for info :star2:


xyzz:
it's failing to build due to missing secure unlock configuration, can you fix this?

tomsquest:
Hi @xyzz ,

I added these VIAL_UNLOCK lines:

```
#define VIAL_UNLOCK_COMBO_ROWS {0, 0}
#define VIAL_UNLOCK_COMBO_COLS {0, 1}
```

But, I am unsure of those keys. I searched the repo, and found many different key combinations. I don't know how I can find the right column/row IDs for the Murcielago.

xyzz:
See https://get.vial.today/docs/porting-to-vial.html#6-set-up-a-secure-unlock-combination

elagil:
Very nice, I wasn't even aware of Vial! Thanks for porting it over.

tomsquest:
Hi @xyzz ,

Changed the combo keys to the two thumb keys, according to the matrix in `keymap.c`.

PR #89, opened 2021-12-05
0xcb

Conor-Burns:
Synced 0xcb/1337 and 0xcb/static from QMK master and added the VIAL keymaps


xyzz:
fyi - the github check failed; however I've just updated the repository to latest qmk master, so if you rebase on top of the newest commit it should probably work

PR #101, opened 2021-12-30
Add support for KBD67 MKII RGB V2.

marcelmaltry:
Unfortunately both QMK Settings and Combos had to be disabled to save enough space for the firmware to fit. Directory structure (`vial_v2`) is adapted from KBD75 Rev2 (`vial_rev2`).


xyzz:
Can you put this keymap under `keyboards/kbdfans/kbd67/mkiirgb/v2/keymaps` and rename to `vial` since it only works with v2?

PR #105, opened 2022-01-18
Planck rev6 drop

dijidiji:
- Added Vial support for the Drop Planck rev6 including VialRGB lighting
- Enabled QMK RGB matrix to allow for the aforementioned VialRGB to work

Presumably this works for all Planck variants but I'm unable to test anything other than the Drop rev6.


keyboards/planck/keymaps/vial/olkb-planck-rev6.1.json:Zone.Identifier
```diff
@@ -0,0 +1,3 @@
+[ZoneTransfer]
```
xyzz: Shouldn't be in the PR

xyzz (changes requested):
If you've only tested the keymap with `rev6_drop`, then the keymap should go under the `rev6_drop` folder.

keyboards/planck/keymaps/vial/olkb-planck-rev6.1.json:Zone.Identifier
dijidiji: Oops, seems to have snuck in.

dijidiji:
> If you've only tested the keymap with `rev6_drop`, then the keymap should go under the `rev6_drop` folder.

vial keymap folder moved under `rev6_drop`, to be honest I had no idea keymaps could be that specific but it compiles and functions as expected!

PR #109, opened 2022-02-01
Added Pixel Rain and Fractal to VIALRGB effects

HorrorTroll:
Since QMK base been updated after last breaking changes, there are 2 new RGB Matrix effect called PIXEL_RAIN and PIXEL_FRACTAL has been added. But it still not added on VIALRGB list yet, so this pr will help for that.

This will need to merged with this pr: https://github.com/vial-kb/vial-gui/pull/73


xyzz:
You cannot add an effect into the middle of the enum, it will break boards compiled with older versions of vial-qmk.

see the comment: `/* Note - never reorder or remove these, only add new animations at the end */`

HorrorTroll:
Done, moved them on bottom list now. Re-test and no issue, so it should good to go.

PR #118, opened 2022-02-11
Add Vial keymap for Planck EZ Glow

gaycodegal:
# Vial QWERTY layout for Planck EZ Glow

1. Cloned / Modified from Planck Rev6. 
2. Filled in associated Planck EZ keymap, tested in Vial with matrix mode. 
3. Turned on MIDI support because its a primary motivator for switching to Vial.
4. This doesn't work with VIA, only Vial ATM according to my tests.

More comments in the readme


gaycodegal:
ok i've tested it using aseqdump, MIDI for sure works now

Edit: tested with Ardour, it for sure works

PR #134, opened 2022-03-25
Add duckboard with RGB Matrix/VialRGB 

somepin:
Add duckboard macropad with Vial encoders and VialRGB enabled. The firmware also enables more layers and most RGB matrix animations, so QMK settings, tap dance, and combos are disabled. Additionally, the [squeez-o](https://github.com/nullbitsco/squeez-o#squeez-o) script by nullbits is used to reduce firmware size for the OLED animation. Tested and works on my hardware.


keyboards/doodboard/duckboard_r2/duckboard_r2.c
```diff
@@ -15,3 +15,25 @@
  */
 
 #include "duckboard_r2.h"
+
+#ifdef RGB_MATRIX_ENABLE
```
xyzz: Please put this under the vial keymap (e.g. keymap.c) to prevent conflicts with qmk merges

xyzz (changes requested):

PR #139, opened 2022-04-07
Ported ErgoDox EZ to Vial

ifohancroft:
Ported the ErgoDox EZ to Vial.
The port was tested successfully on a Falba Tech ErgoDox.
The lighting hasn't been enabled/ported however, because: 1. My ErgoDox doesn't have lighting. 2. The Via documentation isn't clear on how to enable the lighting for keyboards using the RGB_MATRIX_ENABLE.

P.S. The keys for unlock are: Left Half's Top Left Key + Right Half's Middle Right Most Key


xyzz:
For RGB matrix you should use VialRGB - it is described in https://get.vial.today/docs/lighting.html#rgb-matrix--vialrgb

ifohancroft:
> For RGB matrix you should use VialRGB - it is described in https://get.vial.today/docs/lighting.html#rgb-matrix--vialrgb

Thank you! Should I add the RGB support even if I can't test it?

PR #141, opened 2022-04-09
Vial Support for Cantor

DrewTChrist:
Hello I'd like to add Vial support for the [Cantor](https://github.com/diepala/cantor). I believe I've set everything up correctly. Please advise if there are any changes that are required.


xyzz:
It's missing the default keymap, you should add it so that the build succeeds

DrewTChrist:
I have added the default keymap.

xyzz:
Looks like it's still failing to compile

keyboards/cantor/rules.mk
```diff
@@ -0,0 +1,5 @@
+BOOTMAGIC_ENABLE = yes      # Enable Bootmagic Lite
+SPLIT_KEYBOARD = yes
+SERIAL_DRIVER = usart
+VIA_ENABLE      = yes    # Enable VIA
+VIAL_ENABLE      = yes    # Enable VIAL
```
xyzz: this should go inside a keymap

xyzz (changes requested):

keyboards/cantor/rules.mk
DrewTChrist: Alright I moved those lines into keyboards/cantor/vial/rules.mk

PR #143, opened 2022-04-11
put rev1 files in rev1 folder and add rev2 folder and files for bm40hsrgb

jimerickson:
put rev1 files in rev1 folder and add rev2 folder and files for bm40hsrgb
renamed vial folder to via (contains via keymap)
qmk, tap dance, combo, key override are disabled otherwise firmware is to large.


xyzz:
- there are conflicts in this PR
- please don't move any keymaps other than the vial ones, a change like this (splitting a keyboard into rev1/rev2) should go into qmk first as that keyboard exists there and isn't split into two

PR #145, opened 2022-04-12
bm40hsrgb_v2

jimerickson:
add bm40hsrgb_v2 folder and files. qmk, combo, tap dance and key override are disabled due to firmware size constraints. after testing keyboard is autodetected by vial with an active rgb matrix. hoping this is a better way to do it than last pull request.


jimerickson:
closing this as it won't build in its own folder only when a sub folder of bm60hsrgb. i can't figure it out.

PR #158, opened 2022-05-04
Adding AKIRA-60

EugenePY:
# Description about the changes
Initial commit for AKIRA-60 vial support, I also find out that the vial-qmk doesn't pass the linking stage, it seem it use the memroy outside the memory map, 
```shell
Linking: .build/ekow_akira_default.elf                                                              [ERRORS]
 |
 | /usr/local/Cellar/arm-gcc-bin@8/8-2019-q3-update_1/bin/../lib/gcc/arm-none-eabi/8.3.1/../../../../arm-none-eabi/bin/ld:rules_memory.ld:314 cannot move location counter backwards (from 0000000020002a98 to 0000000020001800)
```
I'm using stm32f042x chip, it seem there are some errors need to be fixed in the chibios fork(linking script), I'm completely fine with the qmk/qmk_firmware's master. Do I need to resolve the linking error in this pull request?


xyzz:
That error means it's running out of memory (RAM), you can try following https://get.vial.today/docs/firmware-size.html to make it fit

EugenePY:
>  That error means it's running out of memory (RAM), you can try following https://get.vial.today/docs/firmware-size.html to make it fit

Hi, thanks for the reply, I tried disable some features in the link you provided, but it still the same linkage error. stm32f042 only have 6k RAM, how much ram will it take to enable vial feature? It seems qmk compilation summary only shows flash usage and my configuration takes about 27k in flash.

```shell
Size after:
   text    data     bss     dec     hex filename
      0   27446       0   27446    6b36 ekow_akira_default.bin

```

I also try to compile the other board using stm32f042 chip (make phrygian/ph100:default) the compilation showing error during linking.

EugenePY:
> > That error means it's running out of memory (RAM), you can try following https://get.vial.today/docs/firmware-size.html to make it fit
> 
> Hi, thanks for the reply, I tried disable some features in the link you provided, but it still the same linkage error. stm32f042 only have 6k RAM, how much ram will it take to enable vial feature? It seems qmk compilation summary only shows flash usage and my configuration takes about 27k in flash.

```shell
Size after:
   text    data     bss     dec     hex filename
      0   27446       0   27446    6b36 ekow_akira_default.bin

```

EugenePY:
Hi, I fixed the linkage error am I able to merge?

keyboards/ekow/akira/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,18 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+
+<<<<<<< HEAD
```
xyzz: looks like a merge error happened here

keyboards/ekow/akira/rules.mk
```diff
@@ -0,0 +1,20 @@
...
+MCU = STM32F042
+
+# Bootloader selection
+BOOTLOADER = stm32-dfu
+
+# Build Options
+# change yes to no to disable
+BOOTMAGIC_ENABLE = yes      # Enable Bootmagic Lite
+MOUSEKEY_ENABLE = no		# Mouse keys
+<<<<<<< HEAD
```
xyzz: same issue here

xyzz (changes requested):

keyboards/ekow/akira/rules.mk
EugenePY: Sorry I missed those parts, thank you for your time.

PR #166, opened 2022-05-12
Enable Vial for the customMK EVO70

garbsclassic:
Enables Vial for the customMK EVO70. It disables per-key backlighting, but retains limited underglow, the Bongo Cat (a must) and splash screens for the OLED, and enables the handy QMK settings menu.


keyboards/custommk/evo70/keymaps/passing-comet/rules.mk
```diff
@@ -0,0 +1,13 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+
+VIAL_INSECURE = yes
```
xyzz: VIAL_INSECURE should not be set

xyzz (changes requested):

keyboards/custommk/evo70/keymaps/passing-comet/rules.mk
garbsclassic: Removed

PR #171, opened 2022-05-27
Add support for ai03 Soyuz numpad

properlypurple:
Added support for 1U layout, and then options for making the +,Return, and 0 keys 2U to match a numpad layout. The via layout supplied with the keyboard implements 1U by default.


keyboards/ai03/soyuz/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,6 @@
+VIAL_INSECURE = no
```
xyzz: Remove this line

xyzz (changes requested):

keyboards/ai03/soyuz/keymaps/vial/rules.mk
properlypurple: I've removed this. Thank you!

PR #178, opened 2022-06-06
Added support for dz60rgb_ansi v2

DC959:
Created a vial keymap that enables vial support on the dz60rgb_ansi v2.

Updated the readme under the dz60rgb_ansi folder to reflect the updated support and outlined support is only tested and working on the v2 board.

Have had to disable a large number of features in order to make it fit on the ATmega32u4 but enabled vialrgb.

Firmware does not have insecure set and has a 3 key unlock that is tested and working correctly.

The firmware is my current daily driver so works well from my testing and the vial configurator seems happy when making changes to the keyboard layouts, layers and rgb.


keyboards/dztech/dz60rgb_ansi/readme.md
```diff
@@ -5,6 +5,8 @@
 A hotswap 60% RGB keyboard.
 
 * Keyboard Maintainer: [moyi4681](https://github.com/moyi4681)
+* Vial Support: [dc959](https://github.com/dc959)
+    > Vial support currently only tested and working on DZ60RGB_ansi v2
```
xyzz: You shouldn't modify the original readme. If the keymap only works for a specific revision of the keyboard, you should put this keymap under that revision i.e. under `keyboards/dztech/dz60rgb_ansi/v2/keymaps`

xyzz (changes requested):

keyboards/dztech/dz60rgb_ansi/readme.md
DC959: I did have a feeling that'd be the case.

I have moved the vial keymap under a keymaps folder in v2 and have reverted the readme commit

PR #186, opened 2022-06-25
Adding vial support to iris keyboards

brujoand:
Not quite sure what I'm doing here, so I'll need some extra eyes. I was able to build the firmware and flash it to my Iris keyboard with a few disclaimers.

Firstly I had to add my keyboard which is keebio/iris/rev6a I'm planning to add that directly in a MR to qmk, this leads me to my second issue though because the vial.json needs a product id. This is different on all the revisions of the iris, does this matter? How would this be solved for multiple revisions sharing the same layout?

And finally I had to disable quite a few features to get the firmware small enough. 

In any case my keyboard is working perfectly with Vial now, so I thought I'd add the changes here.


xyzz:
For keyboards with multiple revisions, the vial keymap should be under the specific revision, in this case that would be `keebio/iris/rev6a/keymaps/vial`.

brujoand:
Gotcha, I'll refactor than then. Should I also just add the rev6a build here?
Since the layout on all the iris revitions are identical though, is there a way to inherit the keymap and just change the product id in each "sub-keymap"?

PR #193, opened 2022-07-06
Add Vial support to Rama Works M10-A macropad

ardent-handsaw:
### Add VIAL support to [Rama Works M10-A](https://rama.works/m10-a) macropad. 

**Note:** Probably doesn't work for M10-B or -C, which use wilba.tech PCBs.

- Switched off some features to get firmware size below limit (see /keyboards/m10a/rules.mk and /keyboards/m10a/keymaps/vial/rules.mk)
- Reduced number of layers to 4 (from 10 in VIA)

- Set number of tap dance entries to 4


keyboards/m10a/rules.mk
```diff
@@ -8,7 +8,7 @@ BOOTLOADER = atmel-dfu
 #   change yes to no to disable
 #
 BOOTMAGIC_ENABLE = no       # Enable Bootmagic Lite
-MOUSEKEY_ENABLE = yes       # Mouse keys
```
xyzz: this change should go into keymaps/vial/rules.mk

xyzz (changes requested):

ardent-handsaw:
Some extra files snuck into the second commit so I'm closing this while I clean them up

PR #196, opened 2022-07-10
Make optional encoders optional for basketweave-s

pascalpfeil:
- Make optional encoders optional for basketweave-s
- Prevent `The firmware is too large!` by setting `SPACE_CADET_ENABLE=no`
- Change vial unlock combo in case someone has the encoder on the top left key
- Change bootmagic key in case someone has the encoder on the top left key


pascalpfeil:
I would love to have a solution other than disabling mouse keys - any ideas?
`The firmware is too large! 28748/28672 (76 bytes over)`

pascalpfeil:
Opted to go with `SPACE_CADET_ENABLE=no` for now, suggestions welcome

GerritPlehn:
Cool idea to make the encoders optional via a layout option.
Changing unlock and bootmagic keys seems unnecessary to me since the supported encoder can be pressed as well, but it doesn't hurt anything either so that's fine by me.

What do you think about disabling support for locking switches instead of disabling space cadet? The locking keys seem more niche to me than space cadet mode. 
See https://docs.qmk.fm/#/squeezing_avr?id=configh-settings 
```
#undef LOCKING_SUPPORT_ENABLE
#undef LOCKING_RESYNC_ENABLE
```

GerritPlehn:
Optimizing the vial.json file could also be an option to squeeze out a few bytes. Removing excess floating points and re-positioning the decals could save ~30-40 bytes.

pascalpfeil:
> Changing unlock and bootmagic keys seems unnecessary to me since the supported encoder can be pressed as well, but it doesn't hurt anything either so that's fine by me.

It was so annoying to hold the encoder I have that I though I'd change it.

> Removing excess floating points and re-positioning the decals could save ~30-40 bytes.

Does that mean that the json gets stored in ROM as string? Sounds like huge optimization potential

> What do you think about disabling support for locking switches instead of disabling space cadet? The locking keys seem more niche to me than space cadet mode.

That seems like a better solution thank you!

GerritPlehn:
> It was so annoying to hold the encoder I have that I though I'd change it.

Fair enough, I was just thinking of technical reasons why it wouldn't work with an encoder there :+1: 

> Does that mean that the json gets stored in ROM as string? Sounds like huge optimization potential

At least to some extent. There is some optimization being done, removing colors does not change the ROM size for example. I'm not sure how exactly it is stored, I haven't looked into that yet.

xyzz:
These parts can be safely removed from the JSON:

```
  "name": "Basketweave",
  "vendorId": "0x6B62",
  "productId": "0x6869",
  "lighting": "none",
```

```
      {
        "name": "Basketweave"
      },
```

As well as the colors. That said, it's already compressed and minified, so it probably won't get you that much space.

PR #217, opened 2022-08-04
pico: i2c: align code with chibios, avr, documentation

JohSchneider:
some defines and function signatures did not quite match the
documented ones - which lead to compile issues while building the same
code that uses an i2c-driver based vial-qmk/driver for multiple
platforms: [avr, pico]

Signed-off-by: Johannes Schneider <johschneider@googlemail.com>

### DIFF
--- a/platforms/pico/drivers/i2c_master.c
+++ b/platforms/pico/drivers/i2c_master.c
@@ -7,6 +7,17 @@
 #include "pico/binary_info.h"
 #include "hardware/i2c.h"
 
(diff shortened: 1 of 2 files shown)


JohSchneider:
seems like the CI has some fundamental issues?
```
make: qmk: No such file or directory
make: *** [Makefile:396: mechwild/murphpad:vial] Error 1
ERROR: Cannot run "qmk hello"
```

JohSchneider:
wow - that was fast!
thanks (-:

PR #223, opened 2022-08-17
qmk-upstream rp2040

JohSchneider:
upstream qmk-firmware:develop already has RP2040 support
this branch/PR cherrypicks the relevant commits on-top of/for vial-qmk

if there's a planed rebase/merge with the upstream qmk-firmware, this PR might be unnecessary
... but in the meantime: compiles/works for me ;-)


xyzz:
It will be picked up during the next qmk master branch merge

PR #225, opened 2022-08-18
Make qmk_settings respect AUTO_SHIFT definitions.

riannucci:
Previously when `qmk_settings_reset()` was called, it would always
reset `auto_shift` to 0 (i.e. totally disabled). With this change,
all defined AUTO_SHIFT config macros will instead be used as the
default value.


riannucci:
Definitely open to suggestions on how to make the bit sets consistent.

xyzz:
This looks like it'd cause auto shift to always be enabled when qmk settings are enabled, since Vial sets https://github.com/vial-kb/vial-qmk/blob/955da774eb2ad5d3b0170d480e6de9445b85cc99/builddefs/common_features.mk#L589 in order to pull in the autoshift code.

riannucci:
Ah... that's a bummer. Looking closer I think I can change it to use AUTO_SHIFT_DISABLED_AT_STARTUP from https://github.com/vial-kb/vial-qmk/blob/vial/quantum/process_keycode/process_auto_shift.c#L24 and then set that by default in common_features.mk to keep the previous behavior?

riannucci:
(and then users could un-define it if they want it enabled at startup)

Another thing I could do would be to invert this define? It doesn't seem to be documented anywhere, and clearly has no effect in Vial currently. Looks like this was added in https://github.com/qmk/qmk_firmware/pull/14201... WDYT?

riannucci:
OK, updated this to use AUTO_SHIFT_DISABLED_AT_STARTUP, which shouldn't affect existing vial configs.

PR #232, opened 2022-08-27
KBDfans KBD67 MK2 RGB ISO

GerritPlehn:
Adds support for the PCB compatible with KBD67 lite, Blade65, KBD67v3 
`RGB_MATRIX_ENABLE = no` as space saving measure as others would have heavily compromised on functionality due to the wasteful `lufa-ms` bootloader.


guilleortas:
@GerritPlehn Could you confirm that RGB lighting works for the KBD67 Lite MK II RGB ISO? Sorry to nitpick but it's not clear for me from your message and I'd love confirmation. I understand there's a flash size compromise and some other features would need to be disabled. What I'm asking is whether per-key RGB works even if that means disabling all the amazing VIAL features necessary to make room for it.

GerritPlehn:
> @GerritPlehn Could you confirm that RGB lighting works for the KBD67 Lite MK II RGB ISO? Sorry to nitpick but it's not clear for me from your message and I'd love confirmation. I understand there's a flash size compromise and some other features would need to be disabled. What I'm asking is whether per-key RGB works even if that means disabling all the amazing VIAL features necessary to make room for it.

Hey @guilleortas, yes I'd love to clear this up. From the brief testing I did the lighting does indeed work if you make the required space to enable it. I frankly did not do a huge amount of testing, but I was able to toggle the RGB and change hue/saturation/mode with the default key binds. 

I see that leaving the second keymap layer configured with RGB controls is a bit confusing for those who don't go and read the PR message, so we should probably either change that or leave a note in the readme.

guilleortas:
Thanks a lot for your reply @GerritPlehn, this puts me at ease. Indeed adding a note on the readme about the RGB lighting situation would be nice. Cheers!

guilleortas:
@GerritPlehn One last question: would it be possible to use a different bootloader for this board?
I'm asking because I compiled the default vial keymap and I was concerned by the warning message. How likely is bricking a board when using the LUFA Bootloader?

GerritPlehn:
@guilleortas I really dislike the lufa bootloader and almost bricked it. It is possible to replace it, I did it last weekend and wrote a writeup here https://github.com/GerritPlehn/kbdfans-lufa-atmel

guilleortas:
@GerritPlehn Wow that is so cool, great timing too. Thank you so much for documenting it.
Does the keyboard PCB have spots for attaching the jumper wires? If not, what's the skill level required to solder them? It'd be great if you could attach a couple of pics to your write up showing the connections.
Sorry about all the questions! Cheers

GerritPlehn:
@guilleortas there are through hole pads on the pcb you can easily solder to or use some kind of pins to press in. However one pad was either routed wrong or defective on my unit. You can see that in the pictures in my repo. You can hit me up on discord if you want to discuss this further. fldrmaus#1271

guilleortas:
@GerritPlehn ah that's great, through holes are nice and easy to work with. Sorry I hadn't seen the pics before, very useful. Thank you so much for your cooperation, I'll hit you up should I need any more assistance. However, my unit is in preorder without a concrete ETA yet :'(

PR #253, opened 2022-09-26
Macro count increased

Nisker:
This PR increases the amount of possible macros from 16 to 109 (at 110 it will start to overlap with USER00 keycodes)

With [#135](https://github.com/vial-kb/vial-gui/pull/135) the GUI will support macro_count higher than 32

@alexeyatbluescape from the discord asked to be notified

### DIFF
--- a/quantum/dynamic_keymap.c
+++ b/quantum/dynamic_keymap.c
@@ -39,13 +39,7 @@
 #    define NUM_ENCODERS 0
 #endif
 
-#ifndef DYNAMIC_KEYMAP_LAYER_COUNT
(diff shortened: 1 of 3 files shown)


ghost:
should we add this variable description into docs as well, or docs in the different repo?

Nisker:
> should we add this variable description into docs as well, or docs in the different repo?

I added a a 3rd section to the docs here [#24](https://github.com/vial-kb/vial-kb.github.io/pull/24), that will make it to [get.vial.today](https://get.vial.today/manual/macros.html) when/if merged

ghost:
♥️ amazing! thank you @Nisker

PR #262, opened 2022-10-14
Add mini42

controller-works:
This PR adds support for the mini42 split keyboard by Controller Works. The mini42 uses an RP2040 processor with 16MB of external flash. The keys are direct scan across 42 keys. There is a 128x32 OLED on each half. The halves communicate by serial. The default and via keymaps are tested on production hardware.  This keyboard was recently added to QMK:main.


controller-works:
Excellent! Thank you.








---- On Fri, 14 Oct 2022 17:13:58 -0400 xyzz ***@***.***> wrote ---





Merged https://github.com/vial-kb/vial-qmk/pull/262 into vial.

PR #279, opened 2022-11-03
Add support for the Adafruit Macropad

One-Nub:
Side note: I commented out some RGB matrix effects, as well as bootmagic lite support from the root config.h since I thought they may not be necessary - build-wise there is no issue with having these enabled.


xyzz:
If it still fits in the firmware, don't comment them out.

PR #282, opened 2022-11-09
Added tap dance example compatible with Vial

szzze:


xyzz:
This should go into the documentation repo - https://github.com/vial-kb/vial-kb.github.io/tree/main/docs/docs - otherwise nobody will see it.

szzze:
Fair point, thanks. Added it to https://github.com/vial-kb/vial-kb.github.io/pull/29. Will close here

PR #294, opened 2022-11-28
Add vial keymap for Durgod K320

bclarke123:
Added keymap for K320, may work with K310 but I don't have one to test.  Unlock code is ESC + PAUSE.


keyboards/durgod/boards/DURGOD_STM32_F070/board.mk
```diff
@@ -5,7 +5,7 @@ BOARDSRC = $(CHIBIOS)/os/hal/boards/ST_NUCLEO64_F070RB/board.c
 BOARDINC = $(CHIBIOS)/os/hal/boards/ST_NUCLEO64_F070RB
 
 # Include mcu configuration
-EXTRAINCDIRS = $(BOARD_PATH)/boards/DURGOD_STM32_F070
```
bclarke123, line 8: This appears to be a long standing bug that would stop any Durgod STM32 board from working with VIAL

PR #297, opened 2022-11-30
[Keyboard] Added support for a new keyboard - Ishi 80 Mk0

jasonj2232:
Added support for a new keyboard - Ishi 80 Mk0


xyzz:
Not sure about the difference between the "extra" and regular versions, but if it's two separate boards, then they should have different Vial UIDs.

jasonj2232:
It's the same board in terms of layout. The 'extra' version has support for OLED and 1x Encoder. The regular version doesn't support either. 

This is because those two features require additional pins, more than what is available on a standard pro micro (the switch matrix for this board takes up all 18 available GPIO pins on a standard pro micro), so the extra features work only on controllers like the Elite C with extra pins on the bottom.

The code needs to be different for the extra and regular versions because the pins being used are different. In the regular version rows 4 and 5 are connected to pins D1 and D0. These are SCK and SDA pins - the OLED has to be connected to them and nowhere else. Hence in the extra version these pins are connected to the OLED and rows 4 and 5 are connected to F0 and F1 pins on the bottom of the Elite C.

On the keyboard itself the user can decide whether pins D1 and D0 are connected to the rows or the OLED through the use of solder bridges.

Do I still need to go with separate VIAL UIDs?

xyzz:
> Do I still need to go with separate VIAL UIDs?

yes

PR #317, opened 2023-01-07
Add Vial keymap for converter/adb_usb

ryanlau:
Enable LTO, disabled QMK settings and Key Override, and reduced dynamic layers to 2 in order to reduce firmware size.

Added `matrix_is_on` function to `matrix.c` to get security features working.


keyboards/converter/adb_usb/info.json
```diff
@@ -1,206 +1,12 @@
...
-        {"label":"Right", "x":11.5, "y":5.25},
-        {"label":"Down", "x":12.5, "y":5.25},
-        {"label":"Up", "x":13.5, "y":5.25},
-        {"label":"Num 0", "x":15.5, "y":5.25, "w":2},
-        {"label":"Num .", "x":17.5, "y":5.25},
-        {"label":"Num Enter", "x":18.5, "y":4.25, "h":2}
-      ]
-    }
-  }
+  "layouts": {}
```
lesshonor: All changes to this file should be reverted.
At best they have nothing to do with implementing a Vial keymap; at worst they will actively conflict with changes made in upstream QMK. Removing configuration data from info.json is also a regression.

keyboards/converter/adb_usb/matrix.c
```diff
@@ -249,6 +249,11 @@ void matrix_print(void){
 
 }
 
+bool matrix_is_on(uint8_t row, uint8_t col)
+{
+    return (matrix_get_row(row) & (1<<col));
+}
+
```
lesshonor: There's at least a point to this change, but it's still a ticking time bomb for upstream merges.
Consider placing this function within a separate file that is included by `keymaps/vial/rules.mk`, or just throw it in `keymaps/vial/keymap.c` to avoid including another file.
An explanatory comment would be useful in the event this function is added upstream and ends up breaking the Vial keymap -- then even someone who is unfamiliar with the board would be able to make any necessary changes.

lesshonor (changes requested):

keyboards/converter/adb_usb/matrix.c
ryanlau: addressed in https://github.com/vial-kb/vial-qmk/pull/317/commits/7fe8aeecfedc37078f98bc12e59cf09d339a86f1

keyboards/converter/adb_usb/info.json
ryanlau: addressed in https://github.com/vial-kb/vial-qmk/pull/317/commits/a5f19db97803f6d3a02d24dca57127cedf7e6071

lesshonor (approved):
LGTM

PR #324, opened 2023-01-14
Add vial support for dz60rgb v2

piebug:
Hi there! This PR adds Vial support to the DZ60RGB v2 and v2.1 keyboards (should work for both). I followed the basic steps from the docs:

* Copied over the `vial.json` file for this particular keeb
* Created a `vial_v2` folder with the v2 keymap
* Disabled a bunch of features so the firmware would fit on this wee lil board

I have this working perfectly on my v2.1 right now! It's honestly the best. Anyway, figured I'd open up a PR to share on the off-chance that someone else could use it too!


.gitignore
```diff
@@ -82,6 +82,7 @@ secrets.tar
 # Python things
 __pycache__
 .python-version
+.venv/
```
piebug: Selfishly added this to make things a little cleaner on my end - let me know if I should revert!

xyzz: Open a separate PR with that change

piebug: coolio you got it

piebug: https://github.com/vial-kb/vial-qmk/pull/325

Issue #314, opened 2023-01-04
Hello, I'm the designer of Libra Mini. I want to provide the original firmware file for you to refer to the rocker code


kaiser423:

kaiser423:
https://github.com/M-W-H-185/qmk_firmware/tree/master/keyboards/LM/lmj60

PR #270, opened 2022-10-27
Add firmware for the Libra Mini

jakobkg:
This PR adds firmware for the Libra Mini, a 40% that comes from the manufacturer with a Via enabled firmware but no QMK sources are available for it and Via can only be used by sideloading a .json seemingly originating in the depths of the comments of a youtube video. I've made a completely from-scratch firmware for the Libra Mini in this PR.

Marked as draft PR for the moment as ~only a `vial` keymap is implemented, and~ I have yet to fully implement the joystick that this keyboard has.

PR #334, opened 2023-01-15
Add Libra Mini joystick remapping

somepin:
A continuation of #270 that enables the joystick to be remapped as keys in Vial, modifying the code from #314.


jakobkg:
Thank you for implementing this! I unfortunately didn't find the time for this so it's great to see someone else was able to pick up where I left off and get this working

PR #335, opened 2023-01-15
Port MNK65_STM32 to Vial

LoganSchmalz:
As title. Just ported the MNK65_STM32 to Vial. Unlock requires Escape and Up Arrow on a standard 65% layout.


papacow:
I am having some trouble with building the Vial firmware for the 910 V2 ME PCB with the following files. I got a missing keyboard.json error so I changed the name of info.json to keyboard.json which fixed that issue. However, I still can't figure out the error "UNUSED_PINS in config.h is deprecated and will be removed at a later date"
What can I do to fix this error and successfully compile the firmware.

PR #361, opened 2023-02-09
Add Cyboard's Dactyl Manuform options

RealET:
Add common configuration for Cyboard's Dactyl flex-pcb, and the 9 default layouts we offer.


RealET:
need to reformat

PR #363, opened 2023-02-12
Add Degenpad

ThePanduuh:
Add support for Degenpad PCB for private buy numpad.


xyzz:
it's missing the "default" keymap

ThePanduuh:
Seems like a redundancy with QMK but I will push a default keymap when it's sorted over there. I don't really want them to be different.

PR #367, opened 2023-02-17
Port Keychron V1 to VIAL

adophoxia:
As title. (Follows the same implementation as https://github.com/vial-kb/vial-qmk/pull/327)


SentientUmaru:
Can you make a version for the v1 max? apparently they're different

adophoxia:
> Can you make a version for the v1 max? apparently they're different 

@SentientUmaru 
Elaboration required.

mikeduminy:
I'm also curious about the firmware for the V1 Max because I've seen reports that using the V1 firmware bricks a max board

adophoxia:
> I'm also curious about the firmware for the V1 Max because I've seen reports that using the V1 firmware bricks a max board

Curious about what? If it's asking "can I use this firmware for the V1 Max?", the answer is no because it is built differently between the two boards, so you can't use one on the other. 
@mikeduminy

mikeduminy:
Ah I found it here - https://github.com/Keychron/qmk_firmware/tree/wireless_playground/keyboards/keychron/v1_max

PR #312, opened 2022-12-31
adding vial support to preonic rev3

windhausen:
ported preonic rev3 to vial. Added json file. 
since I am no coder I appreciate your help and suggestions.

PR #374, opened 2023-02-19
Renamed keymap to vial

TweetyDaBird:
Randomly named keymaps with vial support is utterly confusing and non-standard.

Should be kept to naming 'vial'.


lesshonor (changes requested):
#312 indicates this keymap was only intended for `preonic/rev3`, so please move it under a keymap folder for that specific revision.

Based on a quick `diff <(qmk info -kb preonic/rev3_drop -f json) <(qmk info -kb preonic/rev3 -f json)` it's probably good for `rev3` and `rev3_drop`, but it's definitely not usable on anything earlier...

xyzz:
since it's failing to compile, it's probably best to just remove it altogether

TweetyDaBird:
I agree, but how is that done easiest?

lesshonor:
> since it's failing to compile, it's probably best to just remove it altogether

It compiles on `preonic/rev3`, as the original PR intended.

> I agree, but how is that done easiest?

See #376 for an example.

lesshonor (approved):

PR #393, opened 2023-03-05
Fixing invalid property type in dependabot.yml

an-achronism:
Apparently needs to be an array, but is currently just "CI", which is causing Dependabot check to fail on GitHub. (It's still just the one workflow that's being triggered to run, but it's now inside an array to stop GitHub complaining.)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,7 +2,8 @@ version: 2
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
-    labels: [CI]
+    labels:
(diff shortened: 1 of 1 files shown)


xyzz:
These shouldn't be here anyway, it just keeps coming back from qmk merges...

an-achronism:
> These shouldn't be here anyway, it just keeps coming back from qmk merges...

Oh, right, I'm with you.

Delete instead then? Or merge temporarily to shut Dependabot up short term and then ditch?

PR #396, opened 2023-03-07
Keychron Q10 `VIALRGB_ENABLE = yes` fix

adophoxia:
Fix spelling of `VIALRGB_ENABLE = yes` inside `rules.mk` of `vial` keymap since it instead read as `VIALRGB_ENABLe = yes`. (2nd `e` in `ENABLE` mistakenly being lowercase)


adophoxia:
Having this mistake caused the GUI to error out when launching it. This is to fix that.

PR #407, opened 2023-03-11
Port Keychron Q2 to VIAL; Re-attempt

adophoxia:
As title; Re-attempt to #341, now more up to date via upstream `qmk:master` as of the 10th of March, 2023


inoperable:
@adophoxia I really appreciate your efforts on doing the whole keychron Qx line.
I find the default keymap.c and the whole includes from common directory incl. custom keychron mac keys 
is more trouble then useful with vial.

dip switch makes customizing your keymap.c not really intuitive, keychron weird custom callback functions aren't really 
helpful either if you take a look at them - those don't make a lot of sense if you want to use vial and all qmk opts enabled.

also, default eeprom size and wear values are quite constrained - the Qx line MCUs (correct me if I am wrong, but most run on STM32L432KC), tuning a config.h makes more sizable eeprom (for macros/td's/combos) available and you don't need
to stick to 4 layers constrain (which I dont understand) and all of that keychron baggage. 

Maybe I'm just wrong - but a lot of default qmk stuff copy/paste could be better when optimized from keychron mainline

adophoxia:
@inoperable, I appreciate the comment about my commitment for the Q series. I do have some things to say regarding your concerns about how the boards are when it comes to the firmware in the repo.
> I find the default keymap.c and the whole includes from common directory incl. custom keychron mac keys
is more trouble then useful with vial. 

There isn't really much in terms of custom code other than:
```c
#define KC_TASK LGUI(KC_TAB)
#define KC_FLXP LGUI(KC_E)
```
Besides that, all code is pristine as per the [PR Checklist](https://docs.qmk.fm/#/pr_checklist?id=keyboard-prs) that is followed.
> * default (and via) keymaps should be “pristine”
>   * bare minimum to be used as a “clean slate” for another user to develop their own user-specific keymap
>   * standard layouts preferred in these keymaps, if possible
>   * should use [encoder map feature](https://docs.qmk.fm/#/feature_encoders?id=encoder-map), rather than encoder_update_user()
>   * default keymap should not enable VIA – the VIA integration documentation requires a keymap called via

> dip switch makes customizing your keymap.c not really intuitive, keychron weird custom callback functions aren't really
helpful either if you take a look at them - those don't make a lot of sense if you want to use vial and all qmk opts enabled.

That bit may be true, when it comes to the callback functions. The dip switch, I don't believe has any effect on those options and features rather than just switching the default layers for Mac and Win compatibility, which is basically switching the positions of the GUI and Alt keys.

> also, default eeprom size and wear values are quite constrained - the Qx line MCUs (correct me if I am wrong, but most run on STM32L432KC), tuning a config.h makes more sizable eeprom (for macros/td's/combos) available and you don't need
to stick to 4 layers constrain (which I dont understand) and all of that keychron baggage.

You can increase this while configuring the firmware, but there is a catch to this. I don't know if you've read much into emulated EEPROM, but when it comes to these specific STM32L432KBU6s, these have a minimum write endurance of 10K. Having no external/built-in EEPROM is one thing, but emulated EEPROM that's taking from the MCU's flash is another. If you were to try and increase the values from:
```c
/* EEPROM Driver Configuration */
#define WEAR_LEVELING_LOGICAL_SIZE 2048
#define WEAR_LEVELING_BACKING_SIZE (WEAR_LEVELING_LOGICAL_SIZE * 2)
```
You could run into issues from not being able to have the MCU boot properly or even potentially eating up more of the flash size, degrading the write endurance of the MCU. (I'm saying that unless you're the kind of person who needs all of the flash space and flash firmware multiple times in one go, then increasing the values would be impractical and in my opinion, not recommended. A for layers, you can just always make your own, add or remove the ones from the current keymaps. 

Bottom line is, I don't take consideration into what I change the files for when porting to VIAL. Only as to Keychron's spec, which is catering to both, Mac and Win compatibility when it comes to specific functions since I follow the same principle as `"default (and via) keymaps should be “pristine”"` as for the `keychron`/`vial` keymap. I have my own `vial` keymap that's more so based off of `default` since I'm on Windows and have no need for Mac functions.

inoperable:
@adophoxia: **Thank you** for your detailed answer! I  didn't wanted to come over as  critic, I really appreciate the work/time you put into it. 

I experimented with the wear leveling logic, with mixed results :wink: 

Since the MCU has 512KB (at least that what STM32CubeProgrammer says), I thought ramping some values up won't hurt. I overridden most of config.h in my own keymap dir and compiled it with 64KB (max of what QMK can use if I understand the docs properly). 

Firmware compiled, it "_works_" - but when the MCU losses power it forgets all settings and reverts to defaults :facepalm: 

My other assumption seems also be wrong, since I've got a Q6 model, and I thought the MCU is a STM32L432K**C** the STM32CubeProgrammer tells me so at least (might as well be false detection, I used USB and not connected a probe to the MCU)

PR #419, opened 2023-03-21
Port SteamVan rev1 to VIAL

agnulf:
VIAL port for SteamVan rev1.


agnulf:
Mistakenly added the Plaid commit to this pull request.
Will fix and create separate pull requests per keyboard instead.

PR #426, opened 2023-03-25
[Keyboard] Add support for BFO-9000

John-Schreiber:
Add support for BFO-9000


keyboards/keebio/bfo9000/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,6 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+LTO_ENABLE = yes
+TAP_DANCE_ENABLE = no
+COMBO_ENABLE = no
```
lesshonor:
```suggestion
COMMAND_ENABLE = no
```
If you disable `COMMAND` ([which the QMK docs suggest](https://docs.qmk.fm/#/feature_command),  since you can use magic keycodes instead), you can keep these other features with about 1.6kb to spare.

lesshonor (approved):

John-Schreiber:
sounds good to me, Thank you!

PR #259, opened 2022-10-10
VIAL Support for Erkhal's PRKL30 keyboard and my own project, SINPAD20

CountKeepo:
I added VIAL support for the PRKL30 keyboard made by the Finnish Keyboard Community. 
https://github.com/ErkHal/PRKL30

Added all the necessary layouts and keymaps, and also removed a deprecated line from the promicro folder's config.h which kept throwing warnings while compiling the firmware.

Same story for my own little project, the SINPAD20. There's a default, VIA and VIAL keymap.
https://github.com/CountKeepo/sinpad20

PR #209, opened 2022-07-24
Added Johan von Konow 3DPCB Keyboard

jkutianski:
Added port of [Johan von Konow 3DPCB Keyboard](https://github.com/vonkonow/3DPCB-Keyboard) to VIAL

- Support RX/TX leds as NumLock & CapsLock
- Support blink RX/TX leds for layer indication
- Change MO(0) & MO(1) to FN_MO13 & FN_MO23
- Added MIDI support

PR #433, opened 2023-03-29
March merge additions pt2

lesshonor:
With this every valid keyboard† that has a valid `vial` keymap‡ *should* compile (e.g. `qmk mass-compile -km vial`)

Several of these keyboards should probably have distinct `vial` keymaps per-revision, but I'll leave it to someone who actually owns one of the 10 different `keebio/iris` versions to straighten that out.

† `password` and `handwired/3dpcb` don't build at all and I don't care enough to fix them
‡ `handwired/3dpcb` (#209) and `handwired/prkl30` (#259) have `vial` keymaps but no `vial.json` files, so those keymaps don't build and have also never been picked up by CI checks


lesshonor:
I lied; fixed the `default` keymaps for `velocifire/sun20pro` and `ekow/akira` because the failures were irritating.

CountKeepo:
PRKL30 will be fixed with this PR: https://github.com/vial-kb/vial-qmk/pull/434

PR #442, opened 2023-04-19
Add Support for KBDFans D45

Dani-Kirby:
Adding KBDFans D45 and adding Vial functionality


keyboards/kbdfans/d45/d45.c
lesshonor, line 1: This file should be removed (it isn't needed)

keyboards/kbdfans/d45/d45.h
lesshonor, line 1: This file should be removed (it isn't needed)

keyboards/kbdfans/d45/config.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "config_common.h"
+
+/* USB Device descriptor parameter */
+#define DEVICE_VER 0x0001
+#define PRODUCT D45
```
lesshonor: At the very least, `PRODUCT` needs to be quoted:
```suggestion
#define DEVICE_VER 0x0001
#define PRODUCT "D45"
```
In the near future, the USB information needs to be [moved from `config.h` to `info.json`](https://docs.qmk.fm/#/ChangeLog/20221126?id=usb-ids-refactoring). If you don't at least quote the keyboard name, this keyboard will stop compiling when [the changes in this branch](https://github.com/vial-kb/vial-qmk/tree/merge-2023-03-12) are merged into the `vial` branch. That merge may or may not happen before your PR is even accepted.

This general advice to migrate keyboard-level information `info.json` goes for matrix pins, diode direction, MCU and bootloader, and you should apply it to any PR you open that adds keyboards which don't exist upstream. [Please see other examples of how this works in the repository](https://github.com/vial-kb/vial-qmk/blob/vial/keyboards/viktus/minne/info.json). You can also run `qmk info -f json -kb kbdfans/d45/v1` to assist with the conversion (after you move this keyboard to the v1 folder)

keyboards/kbdfans/d45/config.h
```diff
@@ -0,0 +1,51 @@
...
+#define DIODE_DIRECTION COL2ROW
+
+/* Debounce reduces chatter (unintended double-presses) - set 0 if debouncing is not needed */
+#define DEBOUNCE 5
+
+/* Mechanical locking support. Use KC_LCAP, KC_LNUM or KC_LSCR instead in keymap */
+#define LOCKING_SUPPORT_ENABLE
+/* Locking resynchronize hack */
+#define LOCKING_RESYNC_ENABLE
+#define VIAL_KEYBOARD_UID {0xC0, 0x17, 0xF0, 0x89, 0xD6, 0x68, 0xC4, 0xEA}
```
lesshonor: You already have this in `vial/config.h`.
```suggestion
```

lesshonor (changes requested):
You don't need to move the keymaps, but you do need to move the keyboard definition files to `kbdfans/d45/v1`, because `kbdfans/d45/v2` [already exists upstream and is pending a merge into Vial](https://github.com/vial-kb/vial-qmk/tree/merge-2023-03-12/keyboards/kbdfans/d45/v2).

keyboards/kbdfans/d45/config.h
Dani-Kirby: Moved to info.json along with the other bootloader/diode/mcu/lighting/etc.

PR #412, opened 2023-03-14
Port Adelais en Ciel rev3

stanrc85:
Port for Adelais en Ciel rev3 PCB

PR #444, opened 2023-04-19
Add support for mechlovin adelais

Dani-Kirby:
Add support for mechlovin adelais


lesshonor:
You and @stanrc85 should probably have a discussion of some kind.

See also: #408, #412, #432.

stanrc85:
Which version of the PCB and revision is this for? I ended up making Revision specific keymap/vial folders for 2 versions so far to avoid compatibility issues.

Dani-Kirby:
@stanrc85 since you've already done that work on the separate revisions, I'll close this pr out.

PR #447, opened 2023-04-21
Add missing layer `dip switch` code for Keychron V2

adophoxia:
As title.


xyzz:
It fails to compile

adophoxia:
Saw that I forgot to add `#if defined(RGB_MATRIX_ENABLE) && defined(CAPS_LOCK_LED_INDEX)` at the start of the `process_record_kb` function. Compiles fine now.

PR #456, opened 2023-05-08
Create SeijCBoard

seijisandrei:
I want to add a folder for my keyboard project.


xyzz:
there's no keyboard here, just an empty file

PR #457, opened 2023-05-09
Add Drop ALT support

davewongillies:
Its working and I've managed to flash the keyboard but two smallish issues that may or may not matter:

* The keyboard identifies as `Massdrop Inc. ALT Keyboard [VIA]` Vial app but not [vial.rocks](https://vial.rocks). I can't figure out where the [VIA] bit is coming from.
* I followed the porting guide and enabled `vialrgb` mode but using the vial app the lighting tab doesn't show up in the app. It seems to somewhat work on [vial.rocks](https://vial.rocks), but color selection is broken with some errors such as `Uncaught [object ErrorEvent]` and `Uncaught RuntimeError: memory access out of bounds`. Is there anything else I should be enabling? Or is it possible that this keyboard might not be able to fully support vialrgb?

The regular keycodes for changing RGB lighting works though.


xyzz:
> The keyboard identifies as Massdrop Inc. ALT Keyboard [VIA] Vial app but not [vial.rocks](https://vial.rocks/). I can't figure out where the [VIA] bit is coming from.

That happens because the atsam port doesn't support overriding the serial number which is what the desktop Vial application uses to detect Vial keyboards. I recall this line needs to be commented out for a usable Vial support: https://github.com/vial-kb/vial-qmk/blob/a2482e1dc9e26b4053812511c0bf2722c81c6888/tmk_core/protocol/arm_atsam/usb/conf_usb.h#L75

xyzz:
Closing because there's been no update for over a month.

PR #462, opened 2023-05-18
[Keyboard] Add Moondrop Dash75

silvinor:
## Description

Add the Moondrop Dash75 - an innovative 75% keyboard with integrated USB Hub and headphone DAC/Amp

## Upstream

* This ports https://github.com/qmk/qmk_firmware/pull/20890 to **Vial-QMK**
    * and adds the `vail` keymap
    * `via` keymap is forlorn

## Types of Changes
* [X] Keyboard (addition or update)
* [X] Keymap/layout/userspace (addition or update)


keyboards/moondrop/dash75/info.json
```diff
@@ -0,0 +1,119 @@
...
+    "matrix_pins": {
+        "rows": ["B6", "B5", "B4", "D7", "D6", "D4"],
+        "cols": ["C6", "B2", "B3", "F7", "F6", "F5", "F4", "F1", "F0", "B7", "D5", "D3", "D2", "D0", "D1"]
+    },
+    "processor": "atmega32u4",
+    "url": "https://moondroplab.com/en/products/dash75",
+    "usb": {
+        "vid": "0x6DD4",
+        "pid": "0x7501",
+        "device_version": "1.5.0"
```
lesshonor, line 25:
```suggestion
        "device_version": "1.6.0"
```

lesshonor (changes requested):
So, the upstream version of `moondrop/dash75/r1/config.h` has [`#define SERIAL_NUMBER Dash 75`](https://github.com/qmk/qmk_firmware/blob/1aa52124cab7fef299b1031eb527a63de9562667/keyboards/moondrop/dash75/r1/config.h#L6).

...You know this will cause a problem with Vial, which is why it's not in this PR. But it is worth noting that this difference will cause conflicts whenever changes to that file are pulled from upstream, which someone (i.e. xyz) will need to continually monitor and resolve.

I'm not xyz, so maybe he doesn't mind -- but I don't think I'd accept any keymaps to the main repository that are guaranteed to require that kind of babysitting.

keyboards/moondrop/dash75/info.json
silvinor: yeah... the v1.5 for Vial is already avail for download, but this has the incorrect VID/PID.  For the Vial version we increment the version by .1 to inform users who prefer the Vial build over the QMK/VIA build to get this newer one.
*(btw - this is not the only delta, we also needed to remove the `SERIAL_NUMBER` define as Vial uses that to self-identify on the app.)*
I am not aware of any rule that stipulates an exact match to upstream ... if this is so the we'd need to adjust the assumptions on these PR's.

silvinor:
> So, the upstream version of `moondrop/dash75/r1/config.h` has `#define SERIAL_NUMBER Dash 75`.

One thing to consider on this keyboard is that it is targeted to the audiophile consumer - it is constructed with a embedded USB connected to a 4 port-hub, with two USB- Type-A receptacles exposed, then the next two ports have one connected to an embedded MOONRIVER 2 Dual CS43198 DAC, and the other to a rudimentary ATMega32U4 based keyboard. 

The DAC identifies itself with an OEM PID/VID, and a SERIAL NUMBER descriptor as "Moonriver 2".  The s/n on the QMK aligns this branding to the KB too.  Like you said - this does not work with the Vial variant as the App fails to identify the KB as a Vial enabled KB, so was removed for the Vial variant of the code.

image
*(Image extracted from alexotos' review https://youtu.be/6JPqyMEYuwI?t=533 )*

xyzz:
Can you keep the "define" in moondrop/dash75/r1/config.h and then "undef" it in the vial config.h?

silvinor:
> Can you keep the "define" in moondrop/dash75/r1/config.h and then "undef" it in the vial config.h?

Unfortunately not - the sequence of includes means that the `SERIAL_NUMBER` defined in vial code occurs before the inclusion of the kb or keymap `config.h` ... so the define hits an error from the onset.

```bash
Compiling: .build/obj_moondrop_dash75_r1/src/default_keyboard.c                                    In file included from <command-line>:
./keyboards/moondrop/dash75/r1/config.h:6: error: "SERIAL_NUMBER" redefined [-Werror]
 #define SERIAL_NUMBER "Dash 75"
```

I will change the code to use a `#ifndef` ... but that will still be different from QMK upstream anyway.  *(So deleting the def or wrapping it in a `#ifndef` will result in the same outcome.)*

PR #467, opened 2023-05-20
Add Vial support for Zoom65 Lite version

juanlufont:
This PR adds Vial support for the Meletrix Zoom65 Lite version (tested on Olive Dark edition)

To make the firmware binary fit in the MCU memory I have:

- reduced the number of layers
- disabled the *key override* feature
- disabled the *combo* feature

I really do not have any specific reason to choose these features beyond the fact I do not personally use them, so I am open to fine-tune the configuration in any other way if it reduces the size of the firmware enough.


keyboards/meletrix/zoom65_lite/keymaps/vial/keymap.c
```diff
@@ -0,0 +1,92 @@
...
+    encoder_action_unregister();
+    matrix_scan_user();
+}
+
+bool encoder_update_kb(uint8_t index, bool clockwise)
+{
+    encoder_action_register(index, clockwise);
+    return true;
+};
+#endif
```
xyzz: What's that for?

juanlufont:
I am not the author of the functions, they were already presents in the VIA keymap for the Zoom65 Lite included in the upstream QMK repository and they are also present in the qmk-vial fork (I used those files as starting point to create this Vial keymap): https://github.com/vial-kb/vial-qmk/blob/vial/keyboards/meletrix/zoom65_lite/keymaps/via/keymap.c


I am not an expert but they look like handler functions for the encoder events, not sure if provided by Meletrix folks themselves. I just know that the handlers plus my Vial-specific additions work and properly support for the encoder.

keyboards/meletrix/zoom65_lite/keymaps/vial/keymap.c
juanlufont: I am not the author of the functions, they were already presents in the VIA keymap for the Zoom65 Lite included in the upstream QMK repository and they are also present in the qmk-vial fork (I used those files as starting point to create this Vial keymap): https://github.com/vial-kb/vial-qmk/blob/vial/keyboards/meletrix/zoom65_lite/keymaps/via/keymap.c

I am not an expert but they look like handler functions for the encoder events, not sure if provided by Meletrix folks themselves. I just know that the handlers plus my Vial-specific additions work and handle the encoder just right.

lesshonor: yeah, get rid of this chunk. It's a holdover used by VIA from back when VIA didn't support encoders.
just stick with the encoder map

xyzz: You should try deleting that part and check if it still works

juanlufont: I commented the functions and the code still compiles (and the firmware works). I just pushed the change.

PR #472, opened 2023-05-26
Support for ashpil Model M USB-C

GerritPlehn:
Supports both ANSI and ISO layout
Unlocks with Esc + Enter

Thanks to @ashpil for the project.


keyboards/ashpil/modelm_usbc/keymaps/via/keymap.c
lesshonor, line 1: Have you tried compiling this?
Building VIA keymaps isn't supported by the `vial-qmk` repo; you should remove this folder so it won't end up conflicting with a via keymap if it is added to the vanilla `qmk_firmware` repo.

GerritPlehn: I compiled (and tested on hardware) the `vial` keymap without issue, yes. I only added the `via` keymap since the PR comment states
> For keyboard addition or changes, Vial will accept keyboards that implement "default" and "via" keymaps.

lesshonor: That's if you're adding an entire keyboard that doesn't exist upstream...

I'm also pretty sure it was written back when you still could compile VIA keymaps with Vial.

GerritPlehn: ok, thanks for the clarification, I'll remove it in that case.

PR #493, opened 2023-07-12
OLED keymap code for RP2040 and crkbd.

bsas:
Adding custom code to show the keymap in the master OLED (needs RP2040 chips because of size).


bsas:
Vial doesn't accept personal keymaps.

PR #502, opened 2023-07-20
Vial firmware for the mlego m65

pedrorq:
This is a vial firmware for the mlego m65, should be compatible with rev1,2,3,4 and 7.

Fully tested with my rev7 PCB, but it also compiles correctly based on revs 1,2,3,4 original qmk firmwares


lesshonor (commented):
Please revert the changes to files that are in upstream QMK. These will likely cause conflicts in later merges for one...but changes like wiping out the base `info.json` will break all builds for the board generally.

pedrorq:
Should I close this PR, reveert changes and submit a new one so?

PR #503, opened 2023-07-21
First commit

iFIKOZ:


keyboards/namii65/config.h
lesshonor, line 1: You should delete this file as the Vial UID is already set in the Vial keymap's config file.

keyboards/namii65/keymaps/via/config.h
lesshonor, line 1: Rename this folder to `vial`

keyboards/namii65/namii65.h
lesshonor, line 1: Not a requirement, but you should migrate this layout information to `info.json` so it doesn't break in later versions. `qmk migrate -kb namii65`

Then you can delete this file.

lesshonor (changes requested):

PR #519, opened 2023-08-07
Add kbdcraft/adam0110

biuboom-jun:


xyzz:
This keyboard fails to build

PR #525, opened 2023-08-25
Add support for DOIO KB30-01

crkochan:
Adds support for [the DOIO KB30-01 "Megalodon" 30% keyboard from Drop](https://drop.com/buy/megalodon-doio-hot-swappable-rgb-30-keyboard).


crkochan:
Opened #526 with the correct signing key.

PR #528, opened 2023-08-25
horizontal murphpad

phntsm:
Ported default keymap to vial and modified map and json to be oriented horizontally. Custom oled graphic along with custom layer and lock status text. This map has been configured for rp2040 controllers. So far in my testing all seems to be operating as intended. The keymap.c layout is blank save for a qk_boot keycode for flashing adjustment purposes. I wasn't able to get the keymap.c matrix to match the json matrix but all key configuration done in vial matches up as intended on the physical keyboard. Since this is intended for vial anyway, I didn't think it was necessary to bother with it further.


lesshonor:
The main Vial repository doesn't accept personal keymaps.

PR #534, opened 2023-09-05
v1 only works with the SPLIT_USB_DETECT option enabled

nathish:


nathish:
this option is dependent on which MCU you have. closing PR

PR #542, opened 2023-09-12
Add vial port for KBDFans Tet

Shados:
`vial.json` is straight from the via repo keyboards repo, and I've cloned the default keymap. The security unlock keys are present on all possible layouts. Works fine on my Tet.


lesshonor (approved):
Builds fine locally.

PR #544, opened 2023-09-17
Dev void

villeramassana:


lesshonor:
Several of these keymaps/keyboards won't even  compile with the current version of the `vial-qmk` repo. Others would not be accepted with the rules you were supposed to read before opening this PR.
Maybe you should submit one at a time until you understand more about the process.

villeramassana:
Thanks for the info, I’m just trying to make it work but seems to be more
difficult without formation.

Regards

PR #550, opened 2023-09-26
Port MonsGeek M1 to Vial

adophoxia:
As title.


keyboards/monsgeek/m1/keymaps/vial/vial.json
```diff
@@ -0,0 +1,245 @@
+{
+    "name": "MonsGeek M1",
+    "vendorId": "0xFFFE",
+    "productId": "0x0005",
+    "lighting": "vialrgb",
+    "matrix": { "rows": 6, "cols": 16 },
+    "customKeycodes":[
+      {"name":"RESET","title":"RESET EEPROM","shortName":"RESET"}
+    ],
```
lesshonor:
```suggestion
```
Not sure how this got here.

keyboards/monsgeek/m1/keymaps/vial/vial.json
```diff
@@ -0,0 +1,245 @@
...
+    "layouts": {
+      "presets": {"Default": [0,0]},
```
lesshonor:
```suggestion
```
What does this do?

lesshonor (changes requested):

keyboards/monsgeek/m1/keymaps/vial/vial.json
adophoxia: From Xelus via the VIA server:
> presets for the layout options
like default hhkb vs normal 60

lesshonor: What does this do *in Vial*, though?

To cut to the chase: even if it worked, it doesn't seem very useful because the 0th options are already the default.

adophoxia: Probably nothing if I were to hypothesize if the GUI never took this into account. I'll just remove it for convenience's sake.

lesshonor (approved):

PR #551, opened 2023-09-26
Added fullsize ISO layout

omigeot:
I'm a bit new to this, but I needed the ISO layout for IBM(-ish) ModelH. So here's my request, but I haven't been able to find how to compile this particular new layout, so in the end I just switched them in keymap.c ; obviously, I didn't do that in the pull request, so it should be transparent for all ANSI users. Not sure if it needs something else to actually be helpful.


lesshonor:
You probably want to [edit the `vial.json` to add an additional layout](https://get.vial.today/docs/porting-to-via.html#create-layout-options) instead of doing this.

It seems like some passing thought was given to an ISO layout in the current `vial` keymap, but nothing was actually implemented. https://github.com/vial-kb/vial-qmk/blob/ee7f5acefb3fa3e69b1dd2ef325ede1698bef1c9/keyboards/ibm/modelh/keymaps/vial/vial.json#L8

omigeot:
> You probably want to [edit the `vial.json` to add an additional layout](https://get.vial.today/docs/porting-to-via.html#create-layout-options) instead of doing this.

Yeah, it looks there's only an ISO layout in here, and no ANSI one. I do wonder whether there has been a confusion...

lesshonor:
> Yeah, it looks there's only an ISO layout in here, and no ANSI one.

No—
> It seems like some passing thought was given to an ISO layout in the current `vial` keymap, **but nothing was actually implemented**.

You would need to do that "implementing" part. The page I linked to discusses instructions for that. While the instructions for pulling the layout from C header files in QMK are out of date (...you should examine `info.json`, now) the parts about modifying the `vial.json` file and previewing it in the Vial GUI are not. Examining other `vial.json` files in the repository may prove helpful.

xyzz:
Closing because there's been no update for over a month.

PR #558, opened 2023-09-30
Add vial support for Akko Top40

OTLabs:
Vial port for Akko Top40 keyboard,
https://en.akkogear.com/product/acr-top-40/


keyboards/akko/top40/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,15 @@
...
+LTO_ENABLE = yes # make the compiler work harder
+
+# Vial optimization
+# Enable all as the WB32FQ95 has 256kB flash
+
+# QMK optimization
+# Enable all except mostly never used features
+SPACE_CADET_ENABLE = no
+GRAVE_ESC_ENABLE   = no
+MAGIC_ENABLE       = no
```
adophoxia:
```suggestion

```
Having these rules here are redundant due to the WB32FQ95 having 256KB of flash, which is already plenty.

keyboards/akko/top40/keymaps/vial/config.h
```diff
@@ -0,0 +1,20 @@
...
+// Unique keyboard id
+#define VIAL_KEYBOARD_UID {0x5D, 0x72, 0x7C, 0xB4, 0xC9, 0x7C, 0xF2, 0x60}
+
+// Unlock with Escape+Enter
+#define VIAL_UNLOCK_COMBO_ROWS { 0, 1 }
+#define VIAL_UNLOCK_COMBO_COLS { 0, 11 }
+
+// Enable all available RGB Matrix effects
+#define RGB_MATRIX_FRAMEBUFFER_EFFECTS
+#define RGB_MATRIX_KEYPRESSES
```
adophoxia:
```suggestion

```
Why are you reintroducing these rules when 1) These are already defined [at the keyboard level](https://github.com/qmk/qmk_firmware/blob/master/keyboards/akko/top40/config.h#L48-L50) 2) Enabling [`VIALRGB_ENABLE = yes` in the `vial` keymap](https://github.com/vial-kb/vial-qmk/blob/a9f7750316f5e3230c1c3739f7f74b520a0b61ae/keyboards/akko/top40/keymaps/vial/rules.mk#L3) will would actually make every effect be enabled?

keyboards/akko/top40/keymaps/vial/config.h
```diff
@@ -0,0 +1,20 @@
...
+
+// Firmware size optimization
+// disable Lock Key support
+#undef LOCKING_SUPPORT_ENABLE
+#undef LOCKING_RESYNC_ENABLE
```
adophoxia:
```suggestion

```
Any reason this was added?

adophoxia (changes requested):

keyboards/akko/top40/keymaps/vial/rules.mk
OTLabs: If there is plenty of space that does not mean we should waste it, I believe, so I turn on link time optimization for general size optimization and disable the features non-present in this particular keyboard. I would like to insist to leave this options in place even this chip offer a huge flash. What do you think?

keyboards/akko/top40/keymaps/vial/config.h
OTLabs: Mea culpa. I have not checked keyboards level config at first. I had no knowledge about relationship between above mentioned `#define`'s and `VIALRGB_ENABLE = yes`, thank you. Give me a second to fix this.

keyboards/akko/top40/keymaps/vial/rules.mk
OTLabs: Ok, I agree, maybe to leave just LTO.

keyboards/akko/top40/keymaps/vial/config.h
OTLabs: That removes support for nonpresent [Cherry MX Lock switch](https://docs.qmk.fm/#/squeezing_avr?id=configh-settings). I assume that could stay, what do you think?

OTLabs: Ok, let's remove it too just in case some one will get these switches.

adophoxia (approved):

OTLabs:
Thank you for review.

keyboards/akko/top40/keymaps/vial/config.h
lesshonor: > 2) Enabling [`VIALRGB_ENABLE = yes` in the `vial` keymap](https://github.com/vial-kb/vial-qmk/blob/a9f7750316f5e3230c1c3739f7f74b520a0b61ae/keyboards/akko/top40/keymaps/vial/rules.mk#L3) will would actually make every effect be enabled?

This isn't true.
`VIALRGB_ENABLE` is necessary for control of the RGB Matrix feature from the Vial GUI. But effects will not be enabled unless they are explicitly listed in the keyboard or keymap configuration.

lesshonor (approved):

PR #582, opened 2023-10-25
Add Vial support to Keebio Sinc keyboard

LinuxSBC:
This PR adds support for the Keebio Sinc keyboard. I only have the Rev 4 and thus can only test on it, but as it is based on the Via keymap that works for all models, I would assume this also works for all.


keyboards/keebio/sinc/keymaps/vial/keymap.c
```diff
@@ -0,0 +1,42 @@
...
+    _______, _______, _______, _______, _______, _______, _______, _______,          _______, _______, _______, _______, _______, _______, _______, _______
+  ),
+
+  [2] = LAYOUT_all(
+    _______,          _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______,
+    _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______,
+    _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______,
+    _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______,
+    _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______, _______,
+    _______, _______, _______, _______, _______, _______, _______, _______,          _______, _______, _______, _______, _______, _______, _______, _______
```
lesshonor: No need to hardcode filler `KC_TRNS` layers. Dynamic keymaps will automatically fill in "missing" layers at compile-time up to the maximum set in `dynamic_keymap.layer_count`. (By default, this is 4.)
```suggestion
```

keyboards/keebio/sinc/keymaps/vial/keymap.c
```diff
@@ -0,0 +1,42 @@
...
+  )
+
+};
+
+#ifdef ENCODER_MAP_ENABLE
+const uint16_t PROGMEM encoder_map[][NUM_ENCODERS][NUM_DIRECTIONS] = {
+    [0] = { ENCODER_CCW_CW(KC_VOLU, KC_VOLD), ENCODER_CCW_CW(KC_PGUP, KC_PGDN) },
+    [1] = { ENCODER_CCW_CW(RGB_MOD, RGB_RMOD), ENCODER_CCW_CW(KC_MNXT, KC_MPRV) },
+    [2] = { ENCODER_CCW_CW(_______, _______), ENCODER_CCW_CW(_______, _______) },
```
lesshonor: See previous comment.
```suggestion
```

lesshonor (changes requested):
You're going to need to move this under `keebio/sinc/rev4` specifically. The revisions don't have the same matrices, so they can't share a Vial keymap.

...Even if they did, this keymap fails to build for revs 1 and 2. (Please try compiling them next time.)

LinuxSBC:
Thank you for the suggestions. Does this resolve it? Also, I'm sorry for not actually checking to see if it would work on rev1 and rev2. For some reason, I just didn't think of it. Now that I'm considering it more, rev2 and rev1 use a totally different microcontroller, so it would never have worked.

lesshonor (approved):

PR #594, opened 2023-11-08
changed lightning mode of ymdk09 to vial-rgb

choss:
Added the possibility to support the vialrgb on the ymdk09 keypad. Since we had some code space left and this feature is nice, I decided to add it. 

That's all. Nothing more, nothing less


lesshonor (approved):
seems reasonable

keyboards/ymdk/ymd09/keymaps/vial/rules.mk
```diff
@@ -5,3 +5,5 @@ QMK_SETTINGS = no
 TAP_DANCE_ENABLE = no
 COMBO_ENABLE = no
 KEY_OVERRIDE_ENABLE = no
+VIALRGB_ENABLE = yes
+RGB_MATRIX_ENABLE = yes
```
adophoxia:
```suggestion
```
Isn't needed since it's already enabled in `ymd09/info.json`.

adophoxia (changes requested):

keyboards/ymdk/ymd09/keymaps/vial/rules.mk
choss: You are of course right, I've incorporated your comment into the PR.

PR #602, opened 2023-11-15
Porting Inland mk47 from Via

usrfriendly:
This is a quick, effective port from Via to Vial for the Inland MK47 keyboard.  It appears to be fully featured and fully functional with only enough modification to add the required files/configurations Vial needs, simply following the porting guide, compiles small enough to fit on the EEPROM.

PR #607, opened 2023-11-26
Port Drop Alt v2 to VIAL

adophoxia:
As title. 

Since this is the v2 of the original Drop Alt, which uses a ATSAM SAMD51J18A, has (had?) problems with valid licensing, and how as xyzz described in https://github.com/vial-kb/vial-qmk/pull/457#issuecomment-1546782622 regarding overriding the serial number:
> ...the atsam port doesn't support overriding the serial number which is what the desktop Vial application uses to detect Vial keyboards. I recall this line needs to be commented out for a usable Vial support:
> 
> https://github.com/vial-kb/vial-qmk/blob/a2482e1dc9e26b4053812511c0bf2722c81c6888/tmk_core/protocol/arm_atsam/usb/conf_usb.h#L75

So this variant (along with the Ctrl and Shift), that now use an STM32F303, are devoid of these problems.


keyboards/drop/alt/keymaps/vial/vial.json
```diff
@@ -0,0 +1,166 @@
+{
+  "name": "Drop ALT V2",
+  "vendorId": "0x359B",
+  "productId": "0x0006",
```
lesshonor, line 4:
```suggestion
```
These aren't used by Vial.

lesshonor (approved):

keyboards/drop/alt/keymaps/vial/vial.json
adophoxia: Isn't this suggestion a bit different than what you said [last time](https://github.com/vial-kb/vial-qmk/pull/602#discussion_r1395964227)?
> These technically aren't needed for Vial either, but doesn't really matter.

lesshonor: ...sure, I don't always use Ctrl+C Ctrl+V when writing responses

PR #609, opened 2023-11-26
Port Drop Shift v2 to VIAL

adophoxia:
As title. Same as https://github.com/vial-kb/vial-qmk/pull/607


keyboards/drop/shift/keymaps/vial/vial.json
```diff
@@ -0,0 +1,243 @@
+{
+  "name": "Drop SHIFT V2",
+  "vendorId": "0x359B",
+  "productId": "0x000C",
```
lesshonor, line 4:
```suggestion
```
These technically aren't necessary for Vial, but ultimately it doesn't matter much.

lesshonor (approved):

PR #610, opened 2023-11-26
move d45 kbdfans keymap; correct cut-paste error

lesshonor:
Turns out the current Vial keymap for `kbdfans/d45` only works on `kbdfans/d45/v1`, so move it.

Also, correct a cut-paste error.


keyboards/mechwild/puckbuddy/keymaps/vial/vial.json
adophoxia, line 1: There a reason this file was touched?

lesshonor: yes, those are the typos. Look at the description and name vs the shortName; they're reversed.
(It's more like a cut-paste error but I couldn't remember that phrase at the time.)

PR #612, opened 2023-11-30
Adds gpio.h include to ps2_vendor driver

morganvenable:
Core change, just catching up to QMK current..
This enables trackpoint builds on RP2040 to work correctly with Vial by fixing the ps2_vendor driver to include the RP2040 GPIO symbols.

Without including gpio.h, PS2 assignments using pin macros like "GP23" or "GP24" don't work. This include exists in the QMK version of this file already, and works properly there.  So I'm proposing to include it -- but if picking it up directly from QMK is better, that's fine, but it would be nice to fix it now rather than later since it's such a simple touch.


lesshonor:
This PR can now be closed, as these changes have been brought over from upstream QMK.

https://github.com/vial-kb/vial-qmk/blob/b4cfa21cb6bb2b6490268b323eaae8fbc2a19707/platforms/chibios/drivers/vendor/RP/RP2040/ps2_vendor.c#L4

PR #616, opened 2023-12-05
Porting Sofle Choc to Vial.

romones:
Porting Sofle Choc keyboard to Vial from QMK main https://github.com/qmk/qmk_firmware/tree/master/keyboards/sofle_choc 

Following https://get.vial.today/docs/porting-to-vial.html I added a sample KLE generated layout, based config on MX version of the board VIA.

Tested on RP2040 MCU.


xyzz:
the UID check is failing, the Vial UID needs to be unique

romones:
Thanks! I generated the UID and disabled some features to make it fit into an AVR based board.

PR #626, opened 2023-12-20
Add Vault 35 PCB variants: HHKB universal, WKL ortho, WKL universal

josh-l-wang:
Adding three variants of the Vault 35 PCBs. I hope this is okay, as they are based on the same keyboard family and size. Please let me know if you would prefer for these to be separated into individual PRs.


keyboards/jlw/vault35_universal/config.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#define RGB_MATRIX_LED_COUNT 12
+
```
lesshonor, line 4:
```suggestion
```
Don't need to specify this explicitly; QMK will count the number of elements in your `info.json` file.
This applies to all keyboards in this PR.

keyboards/jlw/vault35_universal/keymaps/vial/vial.json
```diff
@@ -0,0 +1,600 @@
+{
+    "name": "Universal Layout Vault35",
```
lesshonor, line 2:
```suggestion
```
This attribute is not used by Vial. This applies to all keyboards in this PR.

lesshonor (approved):

PR #643, opened 2024-01-06
Qh srgb

jsenjobs:


lesshonor:
Per the PR guidelines:

https://github.com/vial-kb/vial-qmk/blob/e49053731e4db354ace3369348f39832693d657e/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L26

It would probably be a good idea to separate out the core changes from the keyboard submission. If you are interested in helping maintain SignalRGB support in Vial going forward, definitely say that too.

xyzz:
I have no idea what most of these changes are for, regarding adding support for stuff like signalrgb it's unlikely to be considered unless you can clearly explain the additional functionality that is not currently implemented on vialrgb side (and even then the priority would be to fix/improve vialrgb instead)

PR #659, opened 2024-01-31
Added vial config for Avalanche keyboard

hcker2000:


hcker2000:
This should pass the build now.

PR #675, opened 2024-02-15
remove advanced PWM timer requirement for complementary outputs

josh-l-wang:
Removing the advanced timer requirement for PWM complementary channel outputs. This has been merged into QMK but not in Vial yet: https://github.com/qmk/qmk_firmware/pull/22830


xyzz:
Seems it's been merged into qmk develop but not in master yet, so it will make its way into vial once qmk merges its develop and master together.

Issue #673, opened 2024-02-15
[Bug] GH60 Rev C Vial Incorrect keymap caused hybrid keys


nlnjnj:
Issue

The right top key（usualy as `~`）does not work with Vial firmware when choose “Split Backspace” layout.

Root Cause
The root cause is straightforward: Incorrect keymap JSON

![%84S1WGCOJQ3$TU44~0ZS$Y](https://github.com/vial-kb/vial-qmk/assets/5831940/3c799ece-102a-4138-bd55-8de66cf5beb5)

![5TY`R P%2ZF)KRGEVMO H1O](https://github.com/vial-kb/vial-qmk/assets/5831940/d3c4fda7-1626-4fec-ae46-5089f436964f)

juanlufont:
Hi! I created that `vial.json` with the wrong code for the split backspace, sorry!

I have created a MR to fix it: https://github.com/vial-kb/vial-qmk/pull/679

PR #679, opened 2024-02-17
Fix wrong code for split backspace in GH60 RevC

juanlufont:
This MR fixes the wrong code for the split backspace in GH60 RevC, reported in <https://github.com/vial-kb/vial-qmk/issues/673>


juanlufont:
Sorry for the many formatting changes in `vial.json`, I did not realized my editor ran `yamlfmt` and applied all those.

PR #686, opened 2024-03-01
Lotus

TweetyDaBird:
Added rp2040_ce and minor updates


keyboards/tweetydabird/lotus58/keymaps/vial/rules.mk
```diff
@@ -7,15 +7,15 @@ VIAL_ENABLE = yes
 ENCODER_MAP_ENABLE = yes
 
 # Reduce size on atmega32u4
-ifeq ($(strip $(CONVERT_TO)), )
+ifeq ($(strip $(MCU)), atmega32u4)
```
lesshonor, line 10: Not sure why this change is being reverted. It will result in these features being disabled (and LTO turned on) by default on all Vial builds (see #684).

TweetyDaBird: Mostly because I had a disagreement with the Git client. We are working it out.

xyzz:
Closing as no updates over a month

PR #719, opened 2024-04-15
Fix winry layout

fmunozs:


fmunozs:
This is my attempt to fix a bug I noticed on the right layout for winry315, I am attaching video of the bug, I am also compiling a new version to confirm if it fixes the problem.

As can be seen on the video, when the right layout is selected all the knob buttons show as HOME.


https://github.com/vial-kb/vial-qmk/assets/1041604/c59e1907-01b8-4f69-9793-7446d42579ef

PR #734, opened 2024-05-19
Add ReviungDash47 & Reviungdash52 vial Support

tumler:
Hello!

This is a PR to add support for the new keyboards:
* [ReviungDash47](https://github.com/tumler/ReviungDash)
* ReviungDash52 (Will be added to the above Repo as well)

Support is only being added for vial (as that is what I use and have tested), therefore the `default` and `vial` folders have been added.


xyzz:
the build is failing because the keyboard UID is not unique

tumler:
Updated

PR #746, opened 2024-06-09
Add support for the Keychron C2 Pro V2 White

StandingPadAnimations:
This patch adds support for the Keychron C2 Pro V2 White, ported from the Keychron fork of QMK. ~~The files for V1 and the V2 RGB model have been removed in the porting process however, since I can't test those.~~ EDIT: these are back in


xyzz:
there are a bunch of conflicts in this pr, please don't touch files that upstream qmk uses

StandingPadAnimations:
The moving of the `ansi` folder to `ansi_v1` was to match the Keychron fork of QMK, since there are 2 versions of the C2 Pro (and I'm using V2)

xyzz:
if it's diverged from the upstream qmk then it will be too much effort for vial to maintain conflicts on every qmk merge, and such pr will not be accepted

PR #753, opened 2024-07-04
Removed redundant modelh folder and added deghosting algorithm

jberclaz:
After all IBM keyboards have been reorganized hierarchically, there was a leftover copy of the Model H controller. This PR removes the redundant copy and moves a keymap which was added to the new folder. In addition, it enables deghosting, as this keyboard needs it.


xyzz:
it's failing to build default keymaps now

jberclaz:
> it's failing to build default keymaps now

Fixed

PR #764, opened 2024-08-02
Add support for boards and pcbs from Nachie

nachie:
Adding support for the following boards. Few of these have already had a public GB, rest have had private GBs with intent to have the design either offered for a full GB in the near future or posted as open source.

Syndrome
Syndrome Ortho
Little Writter
Evelyn
Subtext

Also includes changes from Matt Piasecki for OLED and other support for syndrome. Commits merged into single commit for tracking.


xyzz:
please get rid of the changes outside the keyboards folder

nachie:
Those sneaked in, sorry. Removed

PR #769, opened 2024-08-12
Added vial support for For_Science

arko9699:


keyboards/for_science/keyboard.json
```diff
@@ -14,8 +14,8 @@
     },
     "features": {
         "bootmagic": true,
-        "command": true,
```
xyzz: do not modify files outside of the vial keymap that also exist in upstream qmk, this will cause conflicts during merge

xyzz (changes requested):

xyzz:
there are still modifications outside of the vial keymap

arko9699:
I'm sorry for the frankly egregious series of errors I've made. It should hopefully be correct now.

PR #776, opened 2024-08-24
Add VIAL support for tofu65v2/tofu_jr

Blood13:
After that the VIA stops to work. You can use tofu_jr.json from kbdfans official store. VIA repository needs a json v2 PR for  support of the keyboard.


strikestr (approved):

xyzz:
fails to build

Blood13:
Ok, I'll try to fix it today.

keyboards/dztech/tofu/jr/v1/keymaps/vial/vial.json
```diff
@@ -0,0 +1,165 @@
...
+    "rows": 5,
+    "cols": 15
+  },
+  "lighting": "vialrgb",
+  "keycodes": [
+    "qmk_lighting"
+  ],
+  "menus": [
+    "qmk_rgb_matrix"
+  ],
```
adophoxia:
```suggestion

```
Vial does not recognize these objects, so they can just be deleted.

adophoxia (changes requested):

PR #404, opened 2023-03-10
Add Vial support for Keebio Sinc revs 1-3

kevinmost:
Based this off the initial QMK commit that added rev 3 support: https://github.com/qmk/qmk_firmware/tree/42c240f6cfe697744cdf876ace1e7273f43db60d/keyboards/keebio/sinc

Later snapshots of that directory would not work with Vial due to breaking changes in QMK that have not been merged.

Disabled quite a few features in revs 1 and 2 due to size constraints. However, rev 3 uses a RP2040 with much more flash memory, so nothing had to be disabled there.

Tested:
  - `make keebio/sinc/rev1:vial`, `make keebio/sinc/rev2:vial`, and `make keebio/sinc/rev3:vial` all compile successfully
  - Flashed the rev3 firmware to my Sinc rev3 and am able to use Vial

PR #785, opened 2024-09-05
keebio/sinc: add vial support for rev2

ti-mo:
Hi, looks like this was attempted before in https://github.com/vial-kb/vial-qmk/pull/404, though unsuccessfully. A few things have changed in QMK, and the official rev4 vial keymap landing was a huge help in getting this set up.

As such, this is mostly a copy of rev4, though with none of the rp2040 stuff since the rev2 is an atmega that has most of the boilerplate well taken care of by qmk itself. I suspect this can be copy-pasted verbatim to the rev1 board, since afaik it's fairly similar.

The changes made to `keyboard.json` are mostly to switch from rgblight/backlight to rgb_matrix, mainly to support sleep on idle and suspend. While making this PR, I just realized the backlight feature I removed must be for the monochrome per-key lighting but I haven't seen anyone actually use that in the past decade or so. Removing it yielded a nice space savings, but I can try to reenable it if necessary. While I had it enabled during development, I didn't see a way to configure it in the Vial UI, though, so I'm not sure how it's supposed to be used or how it works.

The rest is taken mostly unchanged from the rev4 keymap, with some rgb effects disabled since they don't act on underglow LEDs.

Let me know what you think!


xyzz:
the change to keyboard.json will cause conflicts with qmk, it needs to be limited to the vial folder

xyzz:
there is no update for over a month, closing

PR #794, opened 2024-09-23
Add support for keebio/quefrency revisions 3 and 5

pjones:
The quefrency is a 65% or optionally a 60%.

Thanks!


pjones:
Everything appears to be working on my two keyboards now (rev3 and rev5).

PR #797, opened 2024-10-02
Adding initial support for the EPOMAKER tide75 (Few Issues)…

hrija:
Here's a little PR to add support for the tide75 from EPOMAKER. 

As some of you already know, EPOMAKER's PRs to QMK upstream have been put on hold due to them not releasing their wireless bindings. While these bindings are currently [drafted to be merged for the tide65](https://github.com/qmk/qmk_firmware/pull/24209#issue-2434653482), with the firmware for the tide65 already being in the upstream repo, this is not the case for the tide75 as of yet, so much of the base of this vial implementation has been pulled from [EPOMAKER's own tide75 repo](https://github.com/Epomaker/tide75).

However, this repo seems to be slightly outdated in that there's some bugs that weren't present with the keyboard coming stock. The main issue is that the RGB matrices seem to be messed up, causing per-key RGB effects to be all over the place rather than taking the order they should, as well as the CTRL key not being lit up at all. 

Secondly, the rotary encoder also seems to need to be turned twice to register, which also seems to be an issue with the firmware from EPOMAKER's repo once again. 

To more or less confirm this, I re-flashed both the default and via keymaps which were provided, and the problem persisted, meaning that the official firmware for this board from the [repo](https://github.com/Epomaker/tide75) is buggy, and is different to what came stock with the board. Another giveaway that the firmware was different is that originally, my board's 'manufacturer' was Hangsheng, but since flashing with this firmware has changed to HS.

Currently, I've reached out to EPOMAKER themselves and @sdk66 who seems to be the main firmware developer for the board to see if I can get ahold of the proper, fully-working firmware and implement it here, allowing me to un-draft this PR. Just keep in mind that **flashing this firmware in its current state will mess up your RGB, with no way to go back till Epomaker releases the right firmware**.

PS. I know there's quite a bit of junk within the files that isn't interpreted by vial, I'll be looking to get rid of these later on once I can get ahold of the proper firmware for this board, but if anyone more experienced wants to help me brush everything up I'd appreciate it :D.

This is my first real PR on Github, so apologies if I've done anything wrong :p


JerrettDavis:
Hi @harrythezomby!

I patched the outstanding issues:
- RGB byte ordering (RGB wasn't correct and setting it to BLUE, RED, or GREEN, resulted in the wrong color being displayed)
- RGB Matrix ordering (RGB wasn't linked to the right key matrix, causing effects to render incorrectly)
- Encoder Resolution (No longer requires turning encoder 2 clicks to get an action to occur).

Please feel free to review at your leisure: https://github.com/vial-kb/vial-qmk/commit/9cefba6f6ef58ca35e511b1e34f60a69eae7c1a6

https://github.com/JerrettDavis/vial-qmk/tree/patch-1

xyzz:

PR #803, opened 2024-10-14
New layer for q10. 5 layers now

pcaro:
<!---

    If you are submitting a Vial-enabled keymap for a keyboard in QMK:


pcaro:
An error. Sorry

PR #832, opened 2024-12-23
Added readme file for the rorian04_kbd

Rorian04:
changes
- added readme for rorian04_kbd
- added vial layout file. This has a basic layout to start from, since the code only has the default layer and other layers are just empty layers.
- added ".uf2" file for people who don't want to compile themselves.


xyzz:
these files aren't acceptable in the vial repository

Rorian04:
> these files aren't acceptable in the vial repository

I thought maybe uf2 might not be accepted. So what should I remove just the uf2 file or even the vial layout file is not acceptable

PR #835, opened 2025-01-02
Added support for Idobao id42, TKC m0lly, Checkerboards Quark, and 1upkeyboards pi60_HSE

seirin-blu:
I did a pull request around a month ago that had some issues. It has been long enough since I did that and I wanted to add a new board so I made the changes requested in that one, closed it, then added the id42

- Added vial support for Idobao id42
- Added vial support for TKC m0lly
- Added vial support for Checkerboards Quark
- Added vial support for 1upkeyboards pi60_HSE

As in the previous pull request, it was suggested, but not requested that I change the keymap, I did not do that, as my only intent is to add vial support for these boards.


keyboards/tkc/m0lly/rules.mk
```diff
@@ -0,0 +1 @@
+ENCODER_MAP_ENABLE = yes
```
xyzz: as this is a keyboard from qmk, the changes outside of the vial keymap have to go into qmk, alternatively you can move this change to be under the vial keymap

xyzz (changes requested):

keyboards/tkc/m0lly/rules.mk
seirin-blu: Missed this again, oops. Should be fixed now

xyzz:
fails to build

seirin-blu:
> fails to build

Yep. Pretty close to the limit when compiling with MSYS. I'll trim it down further so its at like 90% instead of like 98% in MSYS

seirin-blu:
Ok, reduced it to 92% inside MSYS. Hopefully it compiles here now

PR #840, opened 2025-01-12
Add memory usage note for ut472

hestela:
ut472 can run into strange issues, such as ctrl key sticking after switching layers. turning off a couple of rgb animations helps with this issue. Adding a readme note in case others run into this issue.


xyzz:
the readme change needs to either go under the vial keymap or into the base qmk

xyzz:

PR #841, opened 2025-01-14
adds support for ploopy mouse

shadow-absorber:
adds vial keymap for ploopy mouse
disable tap dancing as it does not seem feasible on a mouse
enable lto to fit the firmware on the ploopy mouse


xyzz:
the firmware fails to build because it is too large

shadow-absorber:
> the firmware fails to build because it is too large

huh it was working when shadow pushed it... let shadow check again

shadow-absorber:
image
seems to compile fine for shadow and still fit... but could obviously carve out more space if needed

shadow-absorber:
@xyzz disabled key overrides to save space and make sure the firmware can be compiled.... hopefully when the rp2040 model of the ploopy mouse releases it will not have the same problems as this

PR #865, opened 2025-03-09
Add C2 PRO v1/v2

sfjuocekr:
Copied definitions for the C2 PRO v2 from QMK master.
Changed a typo for the C2 PRO v1.
Added VIAL keymaps for both versions.
Generated four UID's.

I only tested this with the v2, I do not own a v1 but I assume it has the same layout.


xyzz:
there are a bunch of conflicts in here now

sfjuocekr:
Fixed, I also reset the RGB effects back to "stock" but I can put those back to make it more complete as I had added ALL the effects to the list one of which I personally use.

That would then revert [cb12238](https://github.com/vial-kb/vial-qmk/pull/865/commits/cb12238d40e808d40841608d1c8516f80deea9bf)

xyzz:
it fails to build

xyzz:

PR #866, opened 2025-03-10
Add Splitkb.com Elora rev1

veypatch:
Add Splitkb.com Elora rev1, which has been merged in to qmk master last week. Keyboard files are the same as qmk master and vial keymap has been added.


veypatch:
Merged latest vial upstream. PR is now vial keymap only.

PR #869, opened 2025-03-25
feat: add vial keymap for zsa/voyager

danielefongo:


danielefongo:
**Note:** I've tested it on my zsa voyager. Everything looks fine.

PR #870, opened 2025-03-25
GCK

Giliy666:
My first experience with VIAL


keyboards/gck/num3x3/keymaps/via/rules.mk
```diff
@@ -0,0 +1,2 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
```
xyzz, line 2: why's there a VIA keymap that defines VIAL_ENABLE = yes?

xyzz:
no response over a month, closing

PR #872, opened 2025-03-30
Add Vial support for keyten/kt60_mu

key10iq:
In addition to usual Vial support, it adds customization of indicator LEDs directly from Vial
photo_2024-07-22_18-28-10
photo_2024-07-22_18-28-38


xyzz:
Since this keyboard comes from qmk, you must not have changes outside of the vial keymap

key10iq:
> Since this keyboard comes from qmk, you must not have changes outside of the vial keymap

Hmm, why not? I already did this: https://github.com/vial-kb/vial-qmk/pull/761
These changes directly affect Vial, as they introduce some necessary settings.
Well, purely in theory, I could first create this PR here, and then in QMK

xyzz:
it causes merge conflicts when qmk changes something and that's extra work for me to merge in the future (which typically I just replace with the version from qmk)

key10iq:
Got it! I'll keep that in mind.
I've reverted everything back to how it was in the QMK repository

PR #873, opened 2025-03-30
Add Vial support for mb-tkl

melonbred:
Adding Vial support for mb-tkl: https://www.melonbred.dev/keyboards/mbtkl

Tested and working on PCB.


melonbred:
Updated "default" and "split" keymaps to remove depreciated RESET keycode.

PR #882, opened 2025-05-10
ported keebio nyquist

ingvardm:
Added support for Keebio Nyquist keyboard with standard 5x12 layout.


xyzz:
is this supposed to be closed in favor of #883 ?

ingvardm:
@xyzz You can accept this one or the other with 4x12 support and close this one. Note that i did not use the layout option for 4x12 layout because it messes with keymap.

xyzz:
we'll take this instead of the other one since the CI cannot build anything other than the vial keymap. however, it currently fails to build

ingvardm:
@xyzz, right, I'll fix the build for rev1-3.
I'll also add 4x12 as layout. The problem with that is that it also should remove the top row in keymap (although the bottom row of matrix is removed) are there an example of how to handle it?

xyzz:
I don't think there's a good solution that doesn't involve custom matrix scan workarounds, it's probably good enough to just make it kill the row and provide some example saved .vil files for your end users to quickly restore the default layout for either 5 or 4 rows when switching between the two

ingvardm:
@xyzz, fixed the build by adding cofigurations per board revision. Please note the reduced feature size.
Also, I cannot test any of the other revisions except rev5, which I own.

xyzz:
the *.vil file should not be committed to the repo

ingvardm:
removed

PR #886, opened 2025-05-12
CXT Studio 12e3 memory optimizations

muge:
As on the tin, https://github.com/clownfish-og/vial-qmk/commit/08550a8 saves an additional 572 bytes by removing some useless/default values in the config.


keyboards/cxt_studio/12e3/keymaps/vial/config.h
```diff
@@ -6,53 +6,7 @@
 #define VIAL_UNLOCK_COMBO_ROWS {0, 0}
 #define VIAL_UNLOCK_COMBO_COLS {2, 3}
```
clownfish-og, line 8:
```suggestion

#define NO_ACTION_ONESHOT
```
saves another 1,208 bytes

keyboards/cxt_studio/12e3/keymaps/vial/config.h
```diff
@@ -6,53 +6,7 @@
...
-#undef RGBLIGHT_EFFECT_CHRISTMAS
-#undef RGBLIGHT_EFFECT_STATIC_GRADIENT
-#undef RGBLIGHT_EFFECT_RGB_TEST
-#undef RGBLIGHT_EFFECT_ALTERNATING
-#undef RGBLIGHT_EFFECT_TWINKLE
-
-#undef ENABLE_RGB_MATRIX_ALPHAS_MODS
-#undef ENABLE_RGB_MATRIX_GRADIENT_UP_DOWN
-#undef ENABLE_RGB_MATRIX_GRADIENT_LEFT_RIGHT
 #undef ENABLE_RGB_MATRIX_BREATHING
```
clownfish-og:
```suggestion
#undef RGB_MATRIX_DEFAULT_MODE
#    define RGB_MATRIX_DEFAULT_MODE RGB_MATRIX_BREATHING
#undef ENABLE_RGB_MATRIX_SOLID_REACTIVE_SIMPLE
```
I can understand you wanting to keep this reactive animation, but for a default vial keymap I would think something less memory-hungry was a better option, and saves an additional 686 bytes

clownfish-og:
Also, for clarity:

> saves an additional 572 bytes by removing some useless/default values in the config

Not quite. You saved space by adding the overrides for the "magic" functions in `keymap.c`. The rest was more for readability and best practices of not duplicating default values and defines.

Also, these are *optional* suggestions and should not be viewed as required for this PR to be merged.

keyboards/cxt_studio/12e3/keymaps/vial/config.h
muge: Gotcha. I figured it was better to have minimal differences between the original QMK version and the VIAL port, but this makes sense.

clownfish-og:
some minor whitespace/EOL things and keymap alignment that could be done, but nothing critical so I'd say LGTM

PR #889, opened 2025-05-18
add dactyl_manuform_pi_pico vial support

GustawXYZ:
Enabling Vial for dactyl_manuform_pi_pico. Keymap is the same as in QMK version. Only added Vial definitions.


keyboards/handwired/dactyl_manuform_pi_pico/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,8 @@
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+
+# VIAL_INSECURE = yes
```
xyzz: remove this line

xyzz (changes requested):

keyboards/handwired/dactyl_manuform_pi_pico/keymaps/vial/rules.mk
GustawXYZ: thanks for the review. done

PR #894, opened 2025-05-28
unlock all features for new rp2040 version of ploopy mouse

shadow-absorber:
removes the tap dance and key override locks from the rp2040 version of ploopy, mouse while keeping them away from the old ATmega32u4 version so both can use as much of vial as possible without memory overflow


keyboards/ploopyco/mouse/rev1_002/rules.mk
```diff
@@ -1,2 +1,9 @@
 # Processor frequency
 F_CPU = 8000000
+
+TAP_DANCE_ENABLE = no
```
xyzz: changes outside of the vial keymap are not allowed, it will conflict with qmk upstream

xyzz (changes requested):

keyboards/ploopyco/mouse/rev1_002/rules.mk
shadow-absorber: ah okay... any other way to make it stick only to the rev1_002 version or should shadow just scrap that part all together??

shadow-absorber: fixed it by moving vial keymaps folders and setting the 2 revisions as their own "keyboards"

keyboards/ploopyco/mouse/rev1_003/keymaps/vial/rules.mk
```diff
@@ -0,0 +1,16 @@
...
+VIA_ENABLE = yes
+VIAL_ENABLE = yes
+
+# TAP_DANCE_ENABLE = no
+# KEY_OVERRIDE_ENABLE = no
+
+
+# LTO_ENABLE = yes
+
+# VIAL_INSECURE = yes
```
xyzz: remove commented config lines here and in the other rules.mk including VIAL_INSECURE

keyboards/ploopyco/mouse/rev1_002/rules.mk
```diff
@@ -1,2 +1,3 @@
 # Processor frequency
 F_CPU = 8000000
+
```
xyzz: revert this change

xyzz (changes requested):

PR #899, opened 2025-06-19
yer

shmokebud:


xyzz:
why are there 99 file changed, which keyboard are you actually adding?

shmokebud:
> why are there 99 file changed, which keyboard are you actually adding?

did not mean to merge that was my bad, still figuring git out

xyzz:

PR #902, opened 2025-06-25
PDF(layer), Caps Word, and Layer Lock for vial-qmk.

getreuer:
As discussed [in this thread](https://github.com/vial-kb/vial-gui/issues/302), this PR adds firmware support for QMK's [Caps Word](https://docs.qmk.fm/features/caps_word) and [Layer Lock](https://docs.qmk.fm/features/layer_lock) to vial-qmk.

The corresponding Vial GUI change is in https://github.com/vial-kb/vial-gui/pull/332

`PDF(layer)` (persistent default layer) is already enabled unconditionally in recent QMK builds, so no firmware-side changes are needed for that.

Details:
* Caps Word and Layer Lock are enabled in `build_vial.mk`. The `CAPS_WORD_INVERT_ON_SHIFT` option is also set. Otherwise, default settings are used.
* The last bits of the `dynamic_vial_get_number_of_entries` message are used to indicate optionally supported features.
* Vial protocol version was already bumped to 6, matching the bump on made on the GUI side corresponding to this change.

Testing: I tested that this change works as expected on my Moonlander.


xyzz:
I should've probably asked earlier but what is the reason to do `-DCAPS_WORD_INVERT_ON_SHIFT` when QMK settings is enabled? these two don't seem like they would be related

getreuer:
Sorry about this, I meant to add `-DCAPS_WORD_INVERT_ON_SHIFT` unconditionally regardless of QMK settings being enabled. In hindsight, I think it should be on line 14 like

```mk
OPT_DEFS += -DVIAL_ENABLE -DNO_DEBUG -DSERIAL_NUMBER=\"vial:f64c2b3c\" -DCAPS_WORD_INVERT_ON_SHIFT
```

getreuer:
I just opened https://github.com/vial-kb/vial-qmk/pull/905 to make that fix.

PR #911, opened 2025-07-13
add vial support for doio kb03

clownfish-og:
add vial keymap for the doio kb03-01 macropad


keyboards/doio/kb03/keymaps/vial/vial.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "KB03-01",
+  "vendorId": "0xD010",
+  "productId": "0x3011",
```
lesshonor:
```suggestion
```
These attributes are not used by Vial.

lesshonor (approved):

keyboards/doio/kb03/keymaps/vial/vial.json
clownfish-og: removed, that PID was not even correct, thanks for the look

PR #918, opened 2025-07-28
feat: adding support for HLB PoorKoi R1/R2

H3lli0n:
* Support for HLB PoorKoi, a 60% stack acrylic based on Waffling60 with custom features and pcb art
* PCB will be strictly the same for round 2 and polycarbonate cnc'd version


xyzz:
default keymap fails to build

H3lli0n:
Fixed default keymap built, sorry for that

PR #921, opened 2025-08-02
viktus-spmini vial support

bombay-duck:
VIAL support for viktus-sp_mini


xyzz:
the vial keymap fails to build

bombay-duck:
> the vial keymap fails to build

Hmm. When I compiled locally, it succeeded `* The firmware size is approaching the maximum - 28230/28672 (98%, 442 bytes free)` but it fails on the CI check with `* The firmware is too large! 29298/28672 (626 bytes over)`

I'll try to reduce it further

keyboards/viktus/sp_mini/keymaps/vial/config.h
```diff
@@ -0,0 +1,31 @@
+#define VIAL_KEYBOARD_UID {0xE5, 0x2B, 0x48, 0x0A, 0x60, 0xE2, 0x5D, 0x31}
+
+#define VIAL_UNLOCK_COMBO_ROWS { 1, 1  }
+#define VIAL_UNLOCK_COMBO_COLS { 2, 2 }
```
xyzz: this is the same key twice, the unlock combo should be at least 2 keys

xyzz (changes requested):

keyboards/viktus/sp_mini/keymaps/vial/config.h
bombay-duck: Updated

PR #929, opened 2025-08-31
Pull in latest from Squalius-cephalus/silakka54

eyesonly:
The vial-qmk version of silakka54:vial keymap is out of date (only has 4 layers, whereas the latest in Squalius-cephalus/silakka54 defines 8 layers inside keymaps/vial/config.h). 

This pull request will pull in the latest from Squalius-cephalus/silakka54 for this keyboard so that Vial can use all 8 layers.


xyzz:
since this keyboard comes from qmk, the changes should be limited to 1) synchronizing changes with the upstream qmk repository 2) changes to the vial keymap

eyesonly:
Agree and that was also my initial thought.

What prompted me to make this PR is that it has changes specifically for Vial:

1. New addition to keyboards/silakka54/keymaps/vial/config.h
#define DYNAMIC_KEYMAP_LAYER_COUNT 8

2. And obviously the corresponding vial/keymap.c defines the extra layers but also gives the chordal hold layout. It looks like the keyboard creator (Squalius-cephalus) only updated the vial keymap for this and not the default keymap.

This is becoming quite a popular keyboard as it can be bought cheaply pre-assembled, so want to give users compiling Vial the best first experience.

xyzz:
no updates in over a month, closing

PR #942, opened 2025-09-21
Add Vial support for keyten/diablo

key10iq:


xyzz:
It fails to build

key10iq:
> It fails to build

There were some issues with the firmware size. I didn't notice because I was compiling it using an older version of the repo
32U4 moment

PR #945, opened 2025-09-25
[ploopyco/trackball] Add vial support

obosob:


xyzz:
it fails to build because the UID is not unique

obosob:
is there any logic or best practice to picking an ID or is it just random?

xyzz:
It should be random, it's explained how to generate one in https://get.vial.today/docs/porting-to-vial.html
the original logic behind Vial UID was that a pcb can only and is able to be flashed with every firmware with the same UID and with no other UID (be it due to differences in hardware layout, MCU selection, bootloader, etc). but that's only really used by vibl and not enforced, we just have a simpler UID unique check

obosob:
Ta, that should be fixed now.

PR #963, opened 2025-11-13
Port Smith and Rune Ironergo keyboard and include QMK kb files

stanrc85:
I ported the Iron Alice (ironergo) to Vial a while ago but it looks like the original QMK files for the board never got added to Vial, so I've included those as well as the necessary Vial files.


stanrc85:
Nevermind, I see the keyboard source was never added to QMK.

PR #970, opened 2025-11-25
Initial commit for Peej/Tripel keyboard VIAL support

2talltech:
I've ported the default keymap to VIAL. The Tripel keyboard is made of 3 PCBs and can accomodate a controller on any one of those 3 - left, middle, and right. This port  works for a keyboard with the controller soldered to the middle PCB. Other configurations are untested. I built mine with an Atmega32u4, so the memory is limited and some QMK features are disabled (repeat keys, combos, and tap dance).


xyzz:
fails to build

PR #980, opened 2025-12-14
Fix: Prevent unintended forced disabling of Auto-Mouse on certain layersFix/layer state set kb

takatu-kazusige:
####Objective
To correct a behavior where the trackball's "Auto-Mouse function" is unintentionally and forcibly disabled when specific layers (Layers 1 and 2) are used for purposes other than the originally intended scroll function (e.g., a Mac environment layer).

####Problem Solved
In the `layer_state_set_kb` function, when layers within the range of 1 to 2 became active, the original design implemented the following fixed controls:

1. Forcibly switching the trackball to scroll mode.
2. **Forcibly disabling the Auto-Mouse function** (`set_auto_mouse_enable`).

This hardcoded behavior caused an unexpected side effect: if a user utilized Layer 1 or 2 for a different keymap, such as **a standard input or OS-optimized key layout (e.g., a Mac layer)**, the Auto-Mouse feature would be forcibly turned OFF the moment they switched to that layer, resulting in an unwelcome interruption of functionality.

####Changes Made
In the `case 1 ... 2` block within the `layer_state_set_kb` function, the following lines related to the forced control of the Auto-Mouse and scroll mode functions have been commented out:

```c
//cocot_set_scroll_mode(true);
//state = remove_auto_mouse_layer(state, false);
//set_auto_mouse_enable(false);
```

As a result, the state of the Auto-Mouse function will now be **preserved (unless changed by the `default` case or user configuration)**, preventing interference with multi-functional layers.

####Potential Side Effects and Mitigation
* **Loss of Original Scroll Functionality:** If Layers 1 and 2 were previously used exclusively for scrolling, this change means layer switching alone will no longer activate scroll mode.
* **Mitigation:** Users who wish to use Layers 1 and 2 for scrolling must now place dedicated keycodes like `SCRL_TO` or `SCRL_MO` in their keymap to manually enable scroll mode.

* **Auto-Mouse Conflict:** When scroll functionality is manually enabled, if the Auto-Mouse feature remains active, there is a potential for conflict between the scroll operation and the button functions triggered by Auto-Mouse.
* **Mitigation:** Users should consider an alternative key operation (e.g., using `AM_TOG` within the scroll layer) to temporarily disable the Auto-Mouse function when using scroll mode.


xyzz:
- the pr has a lot more contents that what you mention in the description
- the change to the chibios-contrib submodule is not acceptable
- please disclose whether you use AI to prepare the contents of this PR

takatu-kazusige:
Closing this PR due to an error in the target repository selection and branch history. I will resubmit the correct fix to the intended repository shortly.

PR #987, opened 2026-01-23
Add support for the Keychron C2 V2 White

StandingPadAnimations:
This adds support for the Keychron C2 V2 White. The RGB version is not supported due to not owning an RGB C2 V2.

One thing this is missing is backlight controls, since the Keychron C2 V2 White uses LED Matrix (https://docs.qmk.fm/features/led_matrix). While this is a feature in upstream QMK, and seems to be here, I'm not sure if Vial explicitly supports LED Matrix in the GUI. LED Matrix is not to be confused with RGB Matrix (the former being with single-color LEDs and not RGB LEDs).

### DIFF
--- a/keyboards/keychron/c2_pro_v2/ansi/white/keymaps/vial/config.h
+++ b/keyboards/keychron/c2_pro_v2/ansi/white/keymaps/vial/config.h
@@ -1,0 +1,7 @@
+/* SPDX-License-Identifier: GPL-2.0-or-later */
+
+#pragma once
+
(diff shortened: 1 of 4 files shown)


PeterMortensen:
Isn't it a C2 Pro V2? As far as know, the [C2](https://www.keychron.com/products/keychron-c2-wired-mechanical-keyboard)  (without "Pro") has proprietary firmware (like the original Keychron K series),

StandingPadAnimations:
> Isn't it a C2 Pro V2? As far as know, the [C2](https://www.keychron.com/products/keychron-c2-wired-mechanical-keyboard) (without "Pro") has proprietary firmware (like the original Keychron K series),

It is a C2 Pro V2, somehow I forgot about the Pro part.

PR #996, opened 2026-02-22
adding a new keyboard "handwired/tnx_corne_f401".

ToYoNiX:
I want to add a new keyboard here as a contribution. It's a handwired corne built with the STM32 black pill dev board and has encoders. I don't think no body did this combination before and that's why I want to contribute. Here is what I did:

I copied the contents of my keyboard code out of qmk repo and into here. Added vial keymap by creating the vial.json file with my layout copied from keyboard layout editor. and that's really it. I hope this helps someone one day and I will try to maintain it as long as I am using this keyboard and it definitely will not be my last one.

BTW, I created a pull request in the official qmk repo and it is not merged yet so I am not sure if it is a requirement so I wanted to mention that.


xyzz:
> BTW, I created a pull request in the official qmk repo and it is not merged yet so I am not sure if it is a requirement so I wanted to mention that.

In that case you should wait for qmk to merge it before opening the Vial pull request otherwise there will be a lot of conflicts on future merges; please reopen a PR when qmk merges it.

PR #1003, opened 2026-03-13
Add RGB_MATRIX_SLEEP definition to config.h

PepinoGz:
Akko 5075: Enable RGB matrix sleep

Added the RGB_MATRIX_SLEEP definition to config.h. This ensures that RGB LEDs turn off automatically when the host computer enters sleep mode or shuts down, preventing unnecessary power consumption and LED wear when the USB ports provide standby power.


xyzz:
- changes must be limited to the vial keymap to ensure no conflicts happen with qmk merges
- please disclose any uses of ai tools

PR #1007, opened 2026-03-23
add mike:raspberry_pico

mikerinderknecht:


xyzz:
- there's no default keymap
- delete the layout.vil file

mikerinderknecht:
Updated with requested changes

PR #1009, opened 2026-04-04
ENABLE_RGB_MATRIX_SOLID_REACTIVE_NEXUS duplicado

eduardomazolini:


xyzz:
it needs to be fixed in qmk or it will cause conflicts during merge

PR #1014, opened 2026-04-28
Keychron K11 Max Vial config with RGB

jakepearson:
Add support for Keychron K11 Max, have been using it for several months and seems totally stable.


xyzz:
there are many changes outside of the keyboard folder, this change cannot be accepted to the vial repository

jakepearson:
Thanks for looking. I'll clean it up.

PR #1018, opened 2026-05-24
add config for Planck rev4

flatrick:
While the Planck rev 4 barely can accommodate QMK with VIAL enabled, by disabling a few things it is possible to make it fit and have a good experience with it. 

As long as you can accept

- not using the little speaker
- not having backlit keys
- not use MAGIC_ keys
- not having QMK Settings
- and having the console disabled

Then this can work really well for you. Personally, I find the ability to use VIAL being more impactful than having any of the functions mentioned above. 

As this is a slimmed down config, I've added a README to ensure that this is the compromise you have to make to get VIAL running on these old keyboards.

Yes, I did use an LLM to accomplish this, but I have manually gone through what it added before committing to git.

With this merge comes a patch for quantum/send_string.c ; on Arch Linux with avr-gcc 15.1.0 at least, I had to add an include for `avr/io.h` for it to compile as `tap_random_base64` needs it.


quantum/send_string/send_string.c
```diff
@@ -24,6 +24,12 @@
 #include "action.h"
 #include "wait.h"
 
+// AVR timer counter registers (TCNT0/1/3/4) used in tap_random_base64 require
```
xyzz, line 27: the changes must be limited to the vial keymap under the keyboard folder

xyzz (changes requested):

PR #1020, opened 2026-06-03
Support Dictation (macOS)

rmpel:
Added key handling for macOS Dictation

I don't know if I'm doing this right, let me know or ignore completely if not :)

(it works for me ;) )

KC_DICTATION with alias KC_DICT and virtual keycode 0xC3

Unfortunately, I don't not know what changes to make so that Vial.rocks actually shows KC_DICT.


xyzz:
this change would have to happen in the base qmk first

PR #1022, opened 2026-06-12
adding support for caps_word, qmk settings tab while enabling the vial_insecure

ankit-vaghela30:
For cornelius keyboard, the vial doesn't have latest features such as QMK settings tab, CAPS_WORD. The challenge is also that we have low memory so enabling all of these features on top of the existing features would put us out of memory. That is why I enabled vial_insecure and also did some other tweaks to fit the essential QMK settings.


xyzz:
VIAL_INSECURE isn't allowed in the repo, and it looks like a pain to maintain in the future with so many overridden settings

PR #1030, opened 2026-07-02
Initialize Quick Tap Term, NKRO, and Grave Escape QMK settings according to config.h options.

getreuer:
This PR makes the initial settings of Quick Tap Term, NKRO, and Grave Escape configurable, following the usual QMK config.h defines for these options.

## Description

After freshly flashing firmware, it would be intuitive if Vial resets the QMK settings struct in such way that fields initialize according to the usual QMK config options. This way, it is predictable how to configure these initial settings in a custom Vial build, simply by defining them in `config.h` in the same way as done in mainline QMK.

For the most part, QMK settings does work this way, but there are a few gaps as [noted in this reddit thread](https://www.reddit.com/r/olkb/s/Z4A5rAJoe8). This PR attempts to improve that. The `qmk_settings_reset()` function is modified to initialize fields to follow these QMK config options:

```.c
// Interval in ms for Quick Tap.
#define QUICK_TAP_TERM <number>  

// Configure NKRO to default to on.
#define NKRO_DEFAULT_ON true

// Grave Escape options:
#define GRAVE_ESC_ALT_OVERRIDE   // Grave Escape always sends Escape if Alt is pressed.
#define GRAVE_ESC_CTRL_OVERRIDE  // Grave Escape always sends Escape if Control is pressed.
#define GRAVE_ESC_GUI_OVERRIDE   // Grave Escape always sends Escape if GUI is pressed.
#define GRAVE_ESC_SHIFT_OVERRIDE // Grave Escape always sends Escape if Shift is pressed.
```

I have tested these changes on a keyboard to verify these config options appear as the initial settings after flashing.

### DIFF
--- a/quantum/qmk_settings.c
+++ b/quantum/qmk_settings.c
@@ -188,7 +188,10 @@ void qmk_settings_init(void) {
 }
 
 void qmk_settings_reset(void) {
-    QS.grave_esc_override = 0;
+    QS.grave_esc_override = (GRAVE_ESC_ALT_OVERRIDE_Defined << 0)
+                          | (GRAVE_ESC_CTRL_OVERRIDE_Defined << 1)
+                          | (GRAVE_ESC_GUI_OVERRIDE_Defined << 2)
+                          | (GRAVE_ESC_SHIFT_OVERRIDE_Defined << 3);
     QS.auto_shift = 0;
     QS.auto_shift_timeout = AUTO_SHIFT_TIMEOUT;
     QS.osk_tap_toggle = ONESHOT_TAP_TOGGLE;
@@ -209,7 +212,7 @@ void qmk_settings_reset(void) {
     QS.combo_term = COMBO_TERM;
     QS.tapping_term = TAPPING_TERM;
     QS.tapping_v2 = 0;
-    QS.quick_tap_term = TAPPING_TERM;
+    QS.quick_tap_term = QUICK_TAP_TERM;
     QS.tap_code_delay = TAP_CODE_DELAY;
     QS.tap_hold_caps_delay = TAP_HOLD_CAPS_DELAY;
     QS.tapping_toggle = TAPPING_TOGGLE;
@@ -220,7 +223,8 @@ void qmk_settings_reset(void) {
     /* must call clear_keyboard for the NKRO setting to not cause stuck keys */
     clear_keyboard();
     keymap_config.raw = 0;
-    keymap_config.oneshot_enable = 1;
+    keymap_config.nkro = NKRO_DEFAULT_ON;
+    keymap_config.oneshot_enable = true;
     eeconfig_update_keymap(&keymap_config);
 
     /* to trigger all callbacks */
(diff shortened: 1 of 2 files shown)


xyzz:
Overall Vial's philosophy for these dynamic settings is to actually rely less on the static configuration, with the idea that a user flashing a Vial firmware onto a Vial-compatible board always ends up in a known default state rather than picking up whatever random configuration the vendor might've put in when copy pasting their keyboard config.

Right now, this is a mixed bag as you can see from some settings being picked up from the config and others not, but I'd say for the future the goal is certainly to rely less on the hardcoded config, not more. Though, I'm open for a discussion on that.

getreuer:
@xyzz Thanks for the response and for explaining the design philosophy! I completely agree that vendors should configure keyboards to start from a predictable "known default state." QMK's own [guidelines for builders](https://docs.qmk.fm/hardware_keyboard_guidelines#keyboard-defaults) echoes this, recommending a "minimal set" of enabled features to avoid confusing new users. It makes sense that Vial wants to enforce that.

Just to share my perspective on where this PR is coming from: there's a contingent of power users who essentially want a "best of both worlds" between Vial and mainline QMK. It's possible to use Vial almost like QMK—customizing `vial/keymap.c`, `config.h`, and `rules.mk`—while keeping Vial's advantage of dynamic keymap editing. Because these users more often recompile and flash their firmware (after tweaking custom C-level behaviors), it is appealing to define certain settings statically. This way, the settings apply immediately after flashing, saving them the step of manually reloading a saved .vil. 

I put this PR together mainly to follow up on [the reddit thread](https://www.reddit.com/r/olkb/s/Z4A5rAJoe8) I mentioned. A user there was confused that Vial didn't pull in their config.h definitions of `QUICK_TAP_TERM` and NKRO like QMK does. Since it's a relatively mild effort to implement (1&ndash;3 lines per option), it seems like a nice convenience to bridge that gap.

That being said, I don't feel super strongly. I totally understand that the larger architectural decisions around static configuration is complex. Whatever you decide is best for the project is fine by me.

bassamanator:
> Overall Vial's philosophy for these dynamic settings is to actually rely less on the static configuration, with the idea that a user flashing a Vial firmware onto a Vial-compatible board always ends up in a known default state rather than picking up whatever random configuration the vendor might've put in when copy pasting their keyboard config.

> Right now, this is a mixed bag as you can see from some settings being picked up from the config and others not, but I'd say for the future the goal is certainly to rely less on the hardcoded config, not more. Though, I'm open for a discussion on that.


I see this as quite problematic, because it disrespects the user.

Here are some of the implications of this philosophy:
- Don't bother customizing your firmware.
- Make sure you download and install a chromium based browser, or the `vial-gui` because we will not respect all the customizations that you put together in the firmware itself, only some of them.

Just because there are vendors out the that don't respect `QMK guidelines for builders`, we all must pay.

I don't see how this helps the user at all.

Highly problematic.

